PR #4, opened 2021-07-20
[ADD] sale_coupon_multi_gift: New module

chienandalu:
This module allows to define multiple reward products on promotions.

To configure multiple product criterias:

#. Go to *Sales > Catalog > Coupon Programs* and select or create a new one.
#. Choose the criteria of appliance you want to use and the minimum quantities, that
   will be used to calculate the times the rewards can be applied.
#. On the *Reward Type* field choose *Multi Gift*.
#. You'll see a now the *Gift list* on which you can add the products you want to give
   away and the quantities for each of them.

An example:

| Qty |  Product  |
|-----|-----------|
|   3 | Product A |
|   2 | Product B |

With a minimum quantity of 3, for every 3 units of products that fulfill the domain
we'd get 3 units of A and 2 of B. So if the valid products quantities are 11, we'd
get 9 of product A and 6 of product B.

cc @Tecnativa TT30850

please review @CarlosRoca13 @pedrobaeza


CarlosRoca13 (commented):
IDK if next error is related with this module, but when I try to access to some created program I'm getting this error 
image

chienandalu:
Fixed @CarlosRoca13

chienandalu:
Ready for review

CarlosRoca13 (approved):

sale_coupon_multi_gift/README.rst
```diff
@@ -0,0 +1,113 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure multiple product criterias:
```
pedrobaeza:
```suggestion
To configure multiple product rewards:
```

sale_coupon_multi_gift/models/sale_coupon_program.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def _compute_order_count(self):
+        """Relay on the order line link for this programas instead of the discount
+        products"""
+        multi_gift_programs = self.filtered(lambda x: x.reward_type == "multi_gift")
+        super(SaleCouponProgram, self - multi_gift_programs)._compute_order_count()
+        for program in multi_gift_programs:
+            orders = (
+                self.env["sale.order.line"]
+                .search(
```
pedrobaeza: Use `read_group` here for performance

sale_coupon_multi_gift/models/sale_coupon_program.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class SaleCouponProgram(models.Model):
+    _inherit = "sale.coupon.program"
+
+    def _compute_order_count(self):
+        """Relay on the order line link for this programas instead of the discount
```
pedrobaeza:
```suggestion
        """Relay on the order line link for these programs instead of the discount
```

sale_coupon_multi_gift/models/sale_coupon_program.py
```diff
@@ -0,0 +1,40 @@
...
+                )
+                .mapped("order_id")
+            )
+            program.order_count = len(orders)
+
+    def action_view_sales_orders(self):
+        res = super().action_view_sales_orders()
+        if self.reward_type != "multi_gift":
+            return res
+        orders = (
```
pedrobaeza, line 29: Cover this with a direct test of the method.

sale_coupon_multi_gift/models/sale_coupon_multi_gift.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class SaleCouponGift(models.Model):
+    _name = "sale.coupon.multi.gift"
```
pedrobaeza: Why not being in the same file than reward one being a o2m child? And the name can also be `sale.coupon.reward.product_line` or similar

sale_coupon_multi_gift/models/sale_coupon_multi_gift.py
```diff
@@ -0,0 +1,21 @@
...
+    _description = "Coupon Multi Gift"
+
+    coupon_reward_id = fields.Many2one(comodel_name="sale.coupon.reward")
+    reward_product_quantity = fields.Integer(
```
pedrobaeza: This shouldn't be the requirement, but the reward quantity. You should also not consider multiple quantities. The gift is rewarded or not, not X times rewarded.

sale_coupon_multi_gift/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id/id,group_id/id,perm_read,perm_write,perm_create,perm_unlink
+access_multi_gift_salesman,salesman,model_sale_coupon_multi_gift,sales_team.group_sale_salesman,1,0,0,0
+access_multi_gift_manager,multi_gift manager,model_sale_coupon_multi_gift,sales_team.group_sale_manager,1,1,1,1
```
pedrobaeza: Is this in line with main coupon tables permissions?

pedrobaeza (changes requested):

sale_coupon_multi_gift/security/ir.model.access.csv
chienandalu: Yes

sale_coupon_multi_gift/models/sale_coupon_multi_gift.py
chienandalu: Changed

sale_coupon_multi_gift/models/sale_coupon_program.py
chienandalu: Done

sale_coupon_multi_gift/models/sale_coupon_program.py
chienandalu: Done

sale_coupon_multi_gift/models/sale_coupon_multi_gift.py
chienandalu: Changes done :+1:

sale_coupon_multi_gift/models/sale_coupon_program.py
```diff
@@ -0,0 +1,38 @@
...
+        for program in multi_gift_programs:
+            orders = self.env["sale.order.line"].read_group(
+                [
+                    ("state", "not in", ["draft", "sent", "cancel"]),
+                    ("coupon_program_id", "=", program.id),
+                ],
+                ["order_id"],
+                ["order_id"],
+            )
+            program.order_count = len(orders)
```
pedrobaeza, line 23: This is not correct. You should group by `coupon_program_id` and `order_id`, with `lazy=False`, and unfold the count.

sale_coupon_multi_gift/models/sale_coupon_reward.py
```diff
@@ -0,0 +1,51 @@
...
+
+class SaleCouponReward(models.Model):
+    _inherit = "sale.coupon.reward"
+
+    coupon_multi_gift_ids = fields.One2many(
+        comodel_name="sale.coupon.reward.product_line",
+        inverse_name="reward_id",
+        string="Gift list",
+    )
+    discount_line_product_id = fields.Many2one(
```
pedrobaeza: Why do you need this?

sale_coupon_multi_gift/models/sale_order.py
```diff
@@ -0,0 +1,157 @@
...
+            lambda x: x.is_reward_line and x.product_id in free_reward_product
+        )
+        return lines | free_reward_product_lines
+
+    def _get_reward_values_multi_gift_line(self, reward_line, program):
+        """Multi Gift reward rules. For every gift reward rule, we'll create a new
+        sale order line flagged as reward line with a 100% discount"""
+        # Compute the proper price_unit to create the new line accordingly.
+        # We use the form class to ensure the proper onchanges.
+        sale_form = Form(self)
```
pedrobaeza: This is going to be very slow when the order contains a lot of lines. I'm thinking on 2 techniques:

1. Create a form from scratch with the same partner
2. Do a `new` sale.order.line and call `onchange` method for the product/quantity fields.

How is this done on core?

sale_coupon_multi_gift/models/sale_order.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def unlink(self):
+        """Avoid unlinking valid multi gift lines since they aren't linked to the
+        discount product of the promotion program"""
+        if not self.env.context.get("valid_multi_gift_lines"):
+            return super().unlink()
+        return super(
+            SaleOrderLine,
+            self.filtered(
+                lambda x: x.id not in self.env.context.get("valid_multi_gift_lines")
```
pedrobaeza, line 156: Is this done the same way on core?

sale_coupon_multi_gift/models/sale_coupon_reward.py
chienandalu: Oops, I don't know where it came from. Removing it

sale_coupon_multi_gift/models/sale_order.py
chienandalu: No, it's explained here: https://github.com/OCA/sale-promotion/blob/a8f67ff597cd8e9d7e58997cda49817b475c9113/sale_coupon_multi_gift/models/sale_order.py#L77-L85

sale_coupon_multi_gift/models/sale_order.py
chienandalu: In core an automatic service product created for each promotion. As the product to enjoy the promotion must be added by the user prior to enjoy the discount we already have the line

https://github.com/odoo/odoo/blob/13.0/addons/sale_coupon/models/sale_order.py#L81-L82

Then they take the easy way:

https://github.com/odoo/odoo/blob/13.0/addons/sale_coupon/models/sale_order.py#L105-L107

sale_coupon_multi_gift/models/sale_coupon_program.py
chienandalu: In this particular case that's no valid, as one promotion unfolds several order lines, so we'd obtain a count equal to the number of rewarded products for each order.

This way, I just count the dicts (one per order) of the result to obtain the count I need (number of orders, not number of rewards)

image

sale_coupon_multi_gift/models/sale_order.py
chienandalu: Change. I took the same approach as in `sale_elaboration` with a new() and a helper method to trigger the onchanges

pedrobaeza (approved):

chienandalu:
It didn't merge...

PR #8, opened 2021-07-26
[ADD] sale_coupon_limit: New module

chienandalu:
Allowes to define a maximum promotions application number per customer

cc @Tecnativa TT30847

please review @sergio-teruel @pedrobaeza


sale_coupon_limit/README.rst
```diff
@@ -0,0 +1,116 @@
...
+#. If you want to constrain the use of the promotion to the salesmen list, set the
+   option *Restrict on listed salesmen* off, so any other salesman won't be able to
+   apply the promotion.
+
+Usage
+=====
+
+Once the program limits are configured, apply the programs as usual in your sale orders.
+
+Once the limit for a cutomer or a salesman is reached, if we try to apply a promotion:
```
pedrobaeza:
```suggestion
Once the limit for a customer or a salesman is reached, if we try to apply a promotion:
```

sale_coupon_limit/README.rst
```diff
@@ -0,0 +1,116 @@
...
+   :local:
+
+Configuration
+=============
+
+To configure customer limits:
+
+#. Go to *Sales > Catalog > Coupon Programs* and select or create a new one.
+#. Set the *Maximum Customer Applications* to the number of times a coupon can be used
+   by a customer.
```
pedrobaeza, line 44:
```suggestion
   by a customer.

NOTE: The customer limit is applied at commercial entity level, not for each contact inside the entity.
```

sale_coupon_limit/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Sale Coupon Limit",
+    "summary": "Allows to define a maximum promotions application number per customer",
```
pedrobaeza:
```suggestion
    "summary": "Restrict number of promotions per customer or salesman",
```

sale_coupon_limit/models/sale_coupon.py
```diff
@@ -0,0 +1,66 @@
...
+        for a detailed case detail."""
+        message = super()._check_coupon_code(order)
+        if message:
+            return message
+        salesman_rule = (
+            order.user_id
+            and self.program_id
+            and self.program_id.rule_salesmen_limit_ids.filtered(
+                lambda x: order.user_id in x.rule_user_ids
+            )[:1]
```
pedrobaeza: No need to get the first if we apply an unique constraint over (program_id, user_id) - and it's very convenient to do it -.

sale_coupon_limit/models/sale_coupon.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def _check_coupon_code(self, order):
+        """Add customer and salesmen limit to program coupons. Check the error strings
+        for a detailed case detail."""
+        message = super()._check_coupon_code(order)
+        if message:
+            return message
+        salesman_rule = (
+            order.user_id
+            and self.program_id
```
pedrobaeza: I think these first 2 conditions are not needed.

sale_coupon_limit/models/sale_coupon.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        # Customer limit rules
+        if self.program_id and self.program_id.rule_max_customer_application:
+            coupons_count = self.search_count(
+                [
+                    ("order_id", "!=", order.id),
+                    (
+                        "sales_order_id.partner_id",
+                        "child_of",
+                        order.partner_id.commercial_partner_id.id,
```
pedrobaeza: This search is not correct if you have a contact in the order. It should be:
```suggestion
                        "sales_order_id.partner_id.commercial_partner_id",
                        "=",
                        order.partner_id.commercial_partner_id.id,
```
Maybe it's interesting to depend on `sale_commercial_partner` for having this field at order level if we want good performance.

sale_coupon_limit/models/sale_coupon.py
```diff
@@ -0,0 +1,66 @@
...
+            if coupons_count >= self.program_id.rule_max_customer_application:
+                message = {
+                    "error": _(
+                        "This promotion was already applied %s times for this "
+                        "customer and there's an stablished limit of %s."
+                    )
+                    % (coupons_count, self.program_id.rule_max_customer_application)
+                }
+        # Salesmen limit rules
+        elif salesman_rule:
```
pedrobaeza: Both should be checked, not one or the other.

sale_coupon_limit/models/sale_coupon.py
```diff
@@ -0,0 +1,66 @@
...
+            coupons_count = self.search_count(
+                [
+                    ("order_id", "!=", order.id),
```
pedrobaeza: For readability, maybe you can have in a variable a base shared domain between both limits, and only add the different part when calling `search_count`.

sale_coupon_limit/models/sale_coupon_program.py
```diff
@@ -0,0 +1,81 @@
...
+from odoo import _, models
+
+
+class SaleCouponProgram(models.Model):
+    _inherit = "sale.coupon.program"
+
+    def _check_promo_code(self, order, coupon_code):
+        """Add customer and salesmen limit to program rules. Check the error strings
+        for a detailed case detail."""
+        message = super()._check_promo_code(order, coupon_code)
```
pedrobaeza, line 12: Aren't you able to share code for the checks given the base params (I understand that the promotion program, the commercial partner and the user), and move such method to that model, and only call them from each place?

sale_coupon_limit/models/sale_coupon_rule_salesmen_limit.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Tecnativa - David Vidal
```
pedrobaeza: Why not putting this definition on the other file as it's a one2many?

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -0,0 +1,26 @@
...
+    )
+    rule_salesmen_limit_ids = fields.One2many(
+        string="Salesmen Limits",
+        comodel_name="sale.coupon.rule.salesmen.limit",
+        inverse_name="rule_id",
+        help="Maximum times a salesman can apply a program. 0 for no limit.",
+    )
+    rule_salesmen_restrict_list_only = fields.Boolean(
+        string="Restrict on listed salesmen",
+        default=True,
```
pedrobaeza: I'm thinking that this default is not sane:

- It will provoke problems on integration tests.
- For existing promotions, it will change its behavior.
- For new created promotions, they won't be applicable until you add the explicit list of salesmen.

So better to put it as False, and maybe an onchange over rule_salesmen_limit_ids that marks it when you add the first salesmen to the list (`len(self.rule_salesmen_limit_ids) == 1`).

sale_coupon_limit/models/sale_coupon_rule_salesmen_limit.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class SaleCouponRuleSalesmenLimit(models.Model):
+    _name = "sale.coupon.rule.salesmen.limit"
```
pedrobaeza: As stated before, unique constraint over `rule_id, user_id`

sale_coupon_limit/models/sale_coupon_rule_salesmen_limit.py
```diff
@@ -0,0 +1,16 @@
...
+    _description = "Coupon Rule Salesmen limits"
+
+    rule_user_ids = fields.Many2many(comodel_name="res.users", string="Salesmen")
```
pedrobaeza: Why many2many? It should be many2one. No problem in adding more lines with the same number, and it makes the performance better with `auto_join=True`.

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -0,0 +1,26 @@
...
+        help="Check it in order to apply restrictions only to the salesmen configured "
+        "in the list.",
+    )
```
pedrobaeza, line 26: This misses a compute field indicating the maximum number of salesmen number that sums up all the lines and being displayed on UI.

sale_coupon_limit/static/description/icon.svg
```diff
@@ -0,0 +1,141 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
pedrobaeza, line 1: Shouldn't be this icon a number and let this icon to incompatibility?

pedrobaeza (changes requested):

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: Precisely the default is the standard behavior. So when is checked, the rules apply to the list only, otherwise they apply to the list and everyone else is banned.

sale_coupon_limit/static/description/icon.svg
chienandalu: The incompatibility one is better :slightly_smiling_face:

sale_coupon_limit/models/sale_coupon_rule.py
pedrobaeza: If you restrict on listed salesmen by default, and no salesmen is put, then you are always forbidding AFAIK, and we don't want that.

chienandalu: Well, it's a naming issue maybe :) The field restricts the salesmen limits, to the list (the rest of users have no limits, as they don't have restrictions). When check, the limits apply to everyone, an a user not in the list can't use that promotion.

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: I don't understand this. What's the purpose of that field?

pedrobaeza: To get the total limit number. It's a requirement of the customer.

sale_coupon_limit/models/sale_coupon_rule.py
pedrobaeza: Please change the name and feature then for making it more comprehensive:
```
rule_salesmen_strict_limit = fields.Boolean(default=False, string="Strict limit", help="If marked, promotion will only allowed for the list of salesmen with their quantities")
```

sale_coupon_limit/models/sale_coupon_rule_salesmen_limit.py
chienandalu: ok :+1:

sale_coupon_limit/models/sale_coupon.py
chienandalu: Ok, with the second, but what if the order has no salesman?

pedrobaeza: Then you have to deal depending if the coupon is restricted to the list.

sale_coupon_limit/models/sale_coupon_program.py
chienandalu: If you look closer, you'll see that every case has a different domain. Anyway I'm refactoring it a little bit so the coupon_code domain part is share between them all

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: done

sale_coupon_limit/models/sale_coupon_rule_salesmen_limit.py
chienandalu: done

sale_coupon_limit/models/sale_coupon.py
chienandalu: done

sale_coupon_limit/models/sale_coupon_rule_salesmen_limit.py
chienandalu: done

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: Ok, take a look now

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -0,0 +1,67 @@
...
+        string="Maximum Customer Applications",
+        default=0,
+        help="Maximum times that a program can be applied to a customer. "
+        "0 for no limit.",
+    )
+    rule_salesmen_limit_ids = fields.One2many(
+        string="Salesmen Limits",
+        comodel_name="sale.coupon.rule.salesmen.limit",
+        inverse_name="rule_id",
+        help="Maximum times a salesman can apply a program. 0 for no limit.",
```
pedrobaeza:
```suggestion
        help="Maximum times a salesman can apply a program. Empty for no limit.",
```

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    rule_salesmen_strict_limit = fields.Boolean(
+        default=False,
+        string="Strict limit",
+        help="If marked, promotion will only allowed for the list of salesmen with "
```
pedrobaeza:
```suggestion
        help="If marked, promotion will only allow for the list of salesmen with "
```

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -0,0 +1,67 @@
...
+        "their quantities",
+    )
+    rule_salesmen_limit_count = fields.Integer(
+        string="Salesmen maximum promotions",
+        compute="_compute_rule_salesmen_limit_count",
+        readonly=True,
```
pedrobaeza: This is the default

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class SaleCouponRuleSalesmenLimit(models.Model):
+    _name = "sale.coupon.rule.salesmen.limit"
+    _description = "Coupon Rule Salesmen limits"
+
+    rule_user_id = fields.Many2one(
+        comodel_name="res.users",
+        string="Salesman",
+        required=True,
```
pedrobaeza: You have put this m2o with the properties of the inverse name (required, ondelete, auto_join), but this is not he one, but `rule_id`.

pedrobaeza (changes requested):
I still no see shared code between `_check_promo_code` and `_check_coupon_code`. Please create an intermediate method that receives the needed args for checking the same conditions and centralize the same code in one place.

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: Changed

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: Fixed

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: Fixed

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: Done

chienandalu:
> I still no see shared code between _check_promo_code and _check_coupon_code. Please create an intermediate method that receives the needed args for checking the same conditions and centralize the same code in one place.

The problem here is that in one side I'm counting different model ocurrences that fall into different related fields in the sale.order. Not much to reuse in this case. In some core cases, there are some simple tests that can be shared, like if the promotion is global and there's another global promotion applied, if the reward is already in the order lines.

pedrobaeza (approved):

victoralmau (approved):
Code review OK.

PR #15, opened 2021-08-27
[FIX] sale_coupon_auto_refresh: fix MissingError error

SilvioC2C:
When applying a promo code via wizard, Odoo creates a sale.order.line *before* filling the sale.order `code_promo_program_id` field.
After the creation, though, the auto refresh override kicks in, calling the `recompute_coupon_lines` method.
That method finds a sale.order.line that was created from a promo program, but such program is not linked to the sale.order, and so the line gets deleted.
However, since the line has just been created, Odoo will try to run the `_check_company` method on it, which will try to read the field `company_id` of the deleted line.
This operation will result in a `MissingError`, raised by the `__get__` method of `Field` class.

This error is fixed by adding the `skip_auto_refresh_coupons` flag to the wizard context when executing the `process_coupon` method.

### DIFF
--- a/sale_coupon_auto_refresh/__init__.py
+++ b/sale_coupon_auto_refresh/__init__.py
@@ -1,1 +1,2 @@
 from . import models
+from . import wizard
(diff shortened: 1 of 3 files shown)


chienandalu (approved):
Looks good :+1:

pedrobaeza (approved):

PR #16, opened 2021-09-10
[IMP] sale_coupon_multi_gift: optional gift

chienandalu:
Now we use the product field in the multi gift line as a mask for a
many2many field that can be used by a third module as a way to offer
optional gifts to the user.

cc @Tecnativa TT30850

@pedrobaeza take a look

### DIFF
--- a/sale_coupon_multi_gift/migrations/13.0.2.0.0/post-migrate.py
+++ b/sale_coupon_multi_gift/migrations/13.0.2.0.0/post-migrate.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.m2o_to_x2m(
+        env.cr,
+        env["sale.coupon.reward.product_line"],
+        "sale_coupon_reward_product_line",
+        "reward_product_ids",
+        openupgrade.get_legacy_name("reward_product_id"),
+    )
--- a/sale_coupon_multi_gift/models/sale_coupon_reward.py
+++ b/sale_coupon_multi_gift/models/sale_coupon_reward.py
@@ -1,6 +1,6 @@
 # Copyright 2021 Tecnativa - David Vidal
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 
 
 class SaleCouponReward(models.Model):
@@ -21,7 +21,10 @@ class SaleCouponReward(models.Model):
                 "Free Products - %s"
                 % (
                     ", ".join(
-                        reward.reward_product_id.name
+                        "{}x {}".format(
+                            reward.reward_product_quantity,
+                            fields.first(reward.reward_product_ids).name,
+                        )
                         for reward in reward.coupon_multi_gift_ids
                     )
                 )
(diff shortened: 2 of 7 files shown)


sale_coupon_multi_gift/migrations/13.0.2.0.0/post-migrate.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.m2o_to_x2m(
+        env.cr,
+        env["sale.coupon.reward.product_line"],
+        "sale_coupon_reward_product_line",
+        "reward_product_ids",
+        "reward_product_id",
```
pedrobaeza: On standard, this column is removed, so you need to previously rename it.

sale_coupon_multi_gift/models/sale_coupon_reward.py
```diff
@@ -38,11 +41,29 @@ class SaleCouponGift(models.Model):
...
+    @api.depends("reward_product_ids")
+    def _compute_reward_default_product_id(self):
+        """This field acts as a cover for a simple many2one behavior of the module.
+        Another module like `sale_coupon_selection_wizard` can dismiss it in order
+        to allow optional"""
+        for line in self:
+            if line.reward_product_ids:
+                line.reward_default_product_id = fields.first(line.reward_product_ids)
+            else:
+                line.reward_default_product_id = False
```
pedrobaeza: AFAIK, this works the same:

```suggestion
            line.reward_default_product_id = fields.first(line.reward_product_ids)
```

sale_coupon_multi_gift/models/sale_order.py
```diff
@@ -12,7 +13,7 @@ def _get_paid_order_lines(self):
         free_reward_product = (
```
pedrobaeza: Plural?

sale_coupon_multi_gift/views/sale_coupon_program_views.xml
```diff
@@ -17,11 +17,8 @@
                     <field name="coupon_multi_gift_ids">
                         <tree editable="top">
                             <field name="reward_product_quantity" string="Qty" />
-                            <field name="reward_product_id" string="Product" />
-                            <field
-                                name="reward_product_uom_id"
-                                groups="uom.group_uom"
-                            />
+                            <field name="reward_default_product_id" string="Product" />
```
pedrobaeza: I think we can put it as readonly here in UI, and maybe call the inverse as the onchange for direct UI reflection.

pedrobaeza (changes requested):

sale_coupon_multi_gift/views/sale_coupon_program_views.xml
chienandalu: Done :+1:

sale_coupon_multi_gift/models/sale_order.py
chienandalu: Yes, it's better. Changed

sale_coupon_multi_gift/models/sale_coupon_reward.py
chienandalu: Nice :+1:

sale_coupon_multi_gift/migrations/13.0.2.0.0/post-migrate.py
chienandalu: Done

chienandalu:
All comments attended

pedrobaeza (approved):

victoralmau (approved):

PR #19, opened 2021-09-22
[FIX] sale_coupon_auto_refresh: crash when removing promotion line

gurneyalex:
we need to pass skip_auto_refresh_coupons=True in the context when
adding / editing / removing lines otherwise we can end up with the
promotion line being added twice or removed twice (the latter causing a
crash)

### DIFF
--- a/sale_coupon_auto_refresh/models/sale_order.py
+++ b/sale_coupon_auto_refresh/models/sale_order.py
@@ -49,19 +49,25 @@ class SaleOrderLine(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        lines = super().create(vals_list)
+        lines = super(
+            SaleOrderLine, self.with_context(skip_auto_refresh_coupons=True)
+        ).create(vals_list)
         lines.mapped("order_id")._auto_refresh_coupons()
         return lines
 
     def write(self, vals):
         orders = self.mapped("order_id")
-        res = super().write(vals)
+        res = super(
+            SaleOrderLine, self.with_context(skip_auto_refresh_coupons=True)
+        ).write(vals)
         orders |= self.mapped("order_id")
         orders._auto_refresh_coupons()
         return res
 
     def unlink(self):
         orders = self.mapped("order_id")
-        res = super().unlink()
+        res = super(
+            SaleOrderLine, self.with_context(skip_auto_refresh_coupons=True)
+        ).unlink()
         orders._auto_refresh_coupons()
         return res


SilvioC2C (approved):
LGTM

damdam-s (approved):
:+1:

sale_coupon_auto_refresh/models/sale_order.py
```diff
@@ -49,19 +49,25 @@ class SaleOrderLine(models.Model):
...
         orders |= self.mapped("order_id")
         orders._auto_refresh_coupons()
         return res
 
     def unlink(self):
         orders = self.mapped("order_id")
-        res = super().unlink()
+        res = super(
+            SaleOrderLine, self.with_context(skip_auto_refresh_coupons=True)
+        ).unlink()
```
chienandalu, line 71: There's no runbot to test it, but what happens when the user deletes the reward line itself. I remember having a recent issue with that.

sale_coupon_auto_refresh/models/sale_order.py
```diff
@@ -49,19 +49,25 @@ class SaleOrderLine(models.Model):
...
         return lines
 
     def write(self, vals):
         orders = self.mapped("order_id")
-        res = super().write(vals)
+        res = super(
+            SaleOrderLine, self.with_context(skip_auto_refresh_coupons=True)
+        ).write(vals)
         orders |= self.mapped("order_id")
         orders._auto_refresh_coupons()
```
chienandalu, line 64: One issue we're experiencing is the performance of this. Maybe we should restrict the recomputations to a list of safe fields, like `product_id`, `product_uom_qty`, `product_uom_id`. `price_unit`, `discount`. This way, we'd avoid unnecessary updates.

The same would go for the sale order.

sale_coupon_auto_refresh/models/sale_order.py
```diff
@@ -49,19 +49,25 @@ class SaleOrderLine(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        lines = super().create(vals_list)
+        lines = super(
+            SaleOrderLine, self.with_context(skip_auto_refresh_coupons=True)
+        ).create(vals_list)
         lines.mapped("order_id")._auto_refresh_coupons()
```
chienandalu, line 55: Another thing to improve the `_auto_refresh_coupons` method would be to disable it whenever the context has the website_id on it, as there's already a `website_sale_coupon` module triggering the change on every `_cart_update`

sale_coupon_auto_refresh/models/sale_order.py
remi-filament: I have just tested it, removing the reward line is causing no issue, it is recreated automatically (I checked it has a different SOL id)

remi-filament:
Thanks @pedrobaeza for linking with my issue, and thanks @gurneyalex for addressing this at the same time I was reporting / testing without success how to overcome this issue.

This PR is an improvement, meaning that it now works with the following actions :
- remove reward line (it is automatically recreated)
- remove reward line + product line at the origin of the reward

However I still get the same error with the following actions :
- deleting the line at the origin of the reward
- reducing quantities of origin line so that reward should be removed (for instance if product < 3)

It seems the issue is raising when reward line should be automatically removed. Only if I remove it manually I do not get an error.

gurneyalex:
@remi-filament I did not check your cases, but it does look like there could be other issues there too. I'll try and dig into this (or maybe get @SilvioC2C to do so because I'm a bit under the water atm).

gurneyalex:
@chienandalu thanks for the feedback on this. I won't be able to fix perf issues in the context of this PR. Could you open an issue and assign it to me, if you too are short on time?

remi-filament:
I checked again the behavior without this module installed, because I thought the issue might be on Odoo side, but it is working OK on Odoo side (except you have to click the Promotion button everytime you change something).

chienandalu:
> @chienandalu thanks for the feedback on this. I won't be able to fix perf issues in the context of this PR. Could you open an issue and assign it to me, if you too are short on time?

Ok, we'll go for further problems later :slightly_smiling_face:

chienandalu (approved):

PR #22, opened 2021-09-28
 [ADD] website_sale_coupon_restrict: New module

chienandalu:
This module allows to configure a coupon program so the validation is restricted to
eCommerce sales.

cc @Tecnativa TT30945

ping @victoralmau @pedrobaeza


pedrobaeza (approved):
It would be good to regenerate README

chienandalu:
Done :+1:

victoralmau (approved):

PR #24, opened 2021-10-05
[14.0][FIX] sale_coupon_auto_refresh: fix MissingError for good

### LINKED ISSUE #20: [14.0] sale_coupon_auto_refresh error when deleting sale order line
After installing sale_coupon_refresh, Promotions are properly applied automatically on sale orders.
However, if I change my mind and update the sale order to remove either promotion line or the sale order line at the origin of the promotion I get a 
MissingError : Missing Record
Record does not exist or has been deleted.

I have been trying to look where the problem stood but could not find the source of the issue...

SilvioC2C:
This commit, alongside with odoo/odoo#78255, fixes MissingError being raised when:
- the line(s) that allow coupons to be applied are deleted
- the line(s) that allow coupons to be applied are updated making coupons applicable no more

Moreover:
- a check on which fields are updated is introduced in order to boost performances
- the module has been slightly refactored for better readability

Closes #20

### DIFF
--- a/sale_coupon_auto_refresh/models/sale_order.py
+++ b/sale_coupon_auto_refresh/models/sale_order.py
@@ -6,7 +6,8 @@ from odoo import api, fields, models
 
 
 class SaleOrder(models.Model):
-    _inherit = "sale.order"
+    _name = "sale.order"
+    _inherit = ["sale.order", "sale.coupon.refresh.mixin"]
 
     # Used in UI to hide the manual button
     auto_refresh_coupon = fields.Boolean(
@@ -15,21 +16,31 @@ class SaleOrder(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        orders = super().create(vals_list)
+        if self._check_skip_refresh():
+            return super().create(vals_list)
+
+        self_ctx = self.with_context(skip_auto_refresh_coupons=True)
+        orders = super(SaleOrder, self_ctx).create(vals_list)
         orders._auto_refresh_coupons()
         return orders
 
     def write(self, vals):
-        res = super().write(vals)
-        self._auto_refresh_coupons()
+        if self._check_skip_refresh():
+            return super().write(vals)
+
+        old_data = self._read_recs_data()
+        self_ctx = self.with_context(skip_auto_refresh_coupons=True)
+        res = super(SaleOrder, self_ctx).write(vals)
+        new_data = self._read_recs_data()
+        if old_data != new_data:
+            self._auto_refresh_coupons()
         return res
 
     def _auto_refresh_coupons(self):
-        if not self.env.context.get("skip_auto_refresh_coupons"):
-            orders = self.filtered(type(self)._allow_recompute_coupon_lines)
-            if orders:
-                orders = orders.with_context(skip_auto_refresh_coupons=True)
-                orders.recompute_coupon_lines()
+        orders = self.filtered(type(self)._allow_recompute_coupon_lines)
+        if orders:
+            orders = orders.with_context(skip_auto_refresh_coupons=True)
+            orders.recompute_coupon_lines()
 
     def _allow_recompute_coupon_lines(self):
         """Returns whether reward lines in order ``self`` can be recomputed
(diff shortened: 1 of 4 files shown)


SilvioC2C:
@gurneyalex @chienandalu @pedrobaeza 

Please feel free to add any comment/suggestion/criticism that you may think of (about this PR in general, and about the fields that should trigger the recomputation in particular).

sale_coupon_auto_refresh/models/sale_order.py
```diff
@@ -5,6 +5,16 @@
...
 
 
+def _get_data_from_triggers(recs, triggers) -> dict:
+    """Retrieves data from records `recs` on fields passed within `triggers`
+
+    :param models.Model recs: records to be read
+    :param set[str] triggers: set of field names
+    :returns: dict mapping each record ID to a list of fields values
+    """
+    return {r.id: [r[t] for t in triggers] for r in recs}
```
simahawk:
```suggestion
    return {r["id"]: r for r in recs.read(triggers, load=False)}
```

sale_coupon_auto_refresh/models/sale_order.py
```diff
@@ -47,27 +79,52 @@ def _allow_recompute_coupon_lines(self):
...
-        res = super(
-            SaleOrderLine, self.with_context(skip_auto_refresh_coupons=True)
-        ).write(vals)
+        old_data = _get_data_from_triggers(self, triggers)
+        self_ctx = self.with_context(skip_auto_refresh_coupons=True)
+        res = super(SaleOrderLine, self_ctx).write(vals)
         orders |= self.mapped("order_id")
-        orders._auto_refresh_coupons()
+        new_data = _get_data_from_triggers(self, triggers)
+        if old_data != new_data:
```
simahawk, line 93: instead of this it would be better IMO to compute an hash and compare the new hash w/ the old one.
You can add a `promotion_coupoin_hash` computed field and then you can simplify everything by triggering the auto refresh only if the hash has changed.

simahawk: Example of a similar approach: https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader_image/models/shopinvader_image_mixin.py#L31

simahawk:
would be nice to have some test to cover the use cases you mentioned ;)

sale_coupon_auto_refresh/models/sale_order_line.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2021 Tecnativa - David Vidal
+# Copyright 2021 Camptocamp - Silvio Gregorini
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class SaleOrderLine(models.Model):
+    _name = "sale.order.line"
```
chienandalu: Why tearing this apart in another file. It's a not written (or maybe it is, I didn't checked :sweat_smile:) convention to have o2m relations attached to its father in the same file. Anyway, for this case it would ease the diff reading keeping it as it was.

sale_coupon_auto_refresh/models/sale_order.py
```diff
@@ -15,21 +16,31 @@ class SaleOrder(models.Model):
 
     @api.model_create_multi
     def create(self, vals_list):
-        orders = super().create(vals_list)
+        if self.env.context.get("skip_auto_refresh_coupons"):
```
chienandalu: One thing that we can check here as well is the context for `website_id` (and there's no need to declare it as a dependency), as `website_sale_coupon` already autorefresh coupons on every `cart_update` and every checkout controller reload, so no need to add an extra load.

chienandalu (commented):
Aside from the nice to have tests already commented. Two little remarks:

edit: by the way, nice job!

sale_coupon_auto_refresh/models/sale_order_line.py
SilvioC2C: I checked the guidelines, which say that the rule is "one model per file" and having multiple models in the same file can be an exception (and it's not mandatory): 

> For models, views and data declarations, split files by the model involved, either created or inherited. [...] An exception can be made when the model is a model intended to be used only as a one2many model nested on the main model.
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#file-naming

I know it's harder to diff files in this PR because of the split, but I found out it was easier to work this way instead of always having to scroll up and down to check if the code was ok on both models. I believe it'll be better this way on the long run.

sale_coupon_auto_refresh/models/sale_order.py
SilvioC2C: @chienandalu will do it now, thanks for the hint!

sale_coupon_auto_refresh/models/sale_order_line.py
pedrobaeza: @SilvioC2C although the guidelines say that is optional to not split, the other guideline is to do the minimal diff and respect the style of previous work, so although you can init a new module and decide to split `sale.order` and `sale.order.line` in 2 files, in this module, that they are already merged, please don't add noise doing such split.

SilvioC2C: @pedrobaeza thanks for the heads-up, I will remove the split then!

chienandalu (approved):
Thanks! :+1:

PR #27, opened 2021-10-21
[FIX] sale_coupon_limit: compatibility with sale_couopon_selection_wizard

chienandalu:
`sale_couopon_selection_wizard` works with new records to probe if a promotion is applicable before apply it for sure. Thus we need to ensure the right id in the domain.

cc @Tecnativa TT32558


pedrobaeza (approved):

pedrobaeza:
Please put a comment before those lines saying that we use `_origin` for being compatible with NewId records.

chienandalu:
Changes done @ernestotejeda please review

ernestotejeda (approved):

Issue #30, opened 2021-10-25
[Question] How to add more units of a gifted product?


alvaro-domatix:
I'm using the sale_coupon_multi_gift module and I have a problem I don't know how to solve.

I have a promo where if a user adds the product A to his cart I gift him the product B (which is a product he can buy in my website). Once he has selected the product A, he will never be able to buy any more units of product B.

How should I proceed to do this?

Thank you.

PR #31, opened 2021-10-25
[FIX] Enable the posibility of adding more products of a gifted product as new line in the order.

alvaro-domatix:
I think this is a posible solution for the question I posed [here](https://github.com/OCA/sale-promotion/issues/30).

This code allows to add as new lines in the order more units of a product that has been gifted for some promotion.


chienandalu (commented):
The fix looks good, but a new module should be made:

- This module doesn't depend on `website_sale_coupon`, so this method isn't available.
- It will happen with `sale_coupon_multiplier_free_product` which uses the same kind of discount lines.

alvaro-domatix:
Alright! I'll do it. Any module name suggestion? @chienandalu

chienandalu:
@pedrobaeza uses to be more inspired with names than me :sweat_smile:

pedrobaeza:
`website_sale_coupon_multi_gift` should be the obvious choose.

chienandalu:
It's not only compatibility with `sale_coupon_multi_gift` but with `sale_coupon_multiplier_free_product` as well (or any other using that kind of rewads)

pedrobaeza:
Uhm, `website_sale_coupon_reward_line_discriminant`? Anyways, shouldn't this be done at `website_sale_coupon` level and it's a bug?

alvaro-domatix:
> Uhm, `website_sale_coupon_reward_line_discriminant`? Anyways, shouldn't this be done at `website_sale_coupon` level and it's a bug?

No because in website_sale_coupon when you gift one product of your shop, it adds one order line for the product (where you can change the quantity) and one for the gift. multi_gift module adds one unique line with 100%discount so you can't modify the quantity as it is a "reward_line".

I will use website_sale_coupon_reward_line_discriminant as the name if you both agree.

chienandalu:
:+1:

website_sale_coupon_reward_line_discriminant/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Website Sale Coupon Reward Line Discriminant",
+    "version": "13.0.1.0.0",
+    "author": "Domatix, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-promotion",
+    "license": "AGPL-3",
+    "category": "Sale",
+    "depends": ["website_sale"],
```
chienandalu:
```suggestion
    "depends": ["website_sale_coupon"],
```

website_sale_coupon_reward_line_discriminant/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Domatix - Álvaro López
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _cart_find_product_line(self, product_id=None, line_id=None, **kwargs):
+        lines = super(SaleOrder, self)._cart_find_product_line(
```
chienandalu:
```suggestion
        lines = super()._cart_find_product_line(
```

website_sale_coupon_reward_line_discriminant/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows to modify quantity of a sale order line that are a product gift reward adding a new line to manage not reward quantities.
```
chienandalu, line 1: Add some background on the modules that would benefit from this technical addons

chienandalu (approved):

website_sale_coupon_reward_line_discriminant/models/sale_order.py
alvaro-domatix: I made all 3 suggested changes. But I would like to ask why this super is better than the other? Or it's just not better, just cleaner?

pedrobaeza: It's a cleaner syntax since Python 3

chienandalu:
@alvaro-domatix Are you still on this?

alvaro-domatix:
Yes, if it is everything fine and I just need to do the test I will do it.

alvaro-domatix:
Hello @chienandalu, I'd like to know if the pre-commit failing has something to do with our code or how to fix it.

alvaro-domatix:
Didn't want to close this PR, can be reopened or do I have to open a new one? @chienandalu @pedrobaeza

alvaro-domatix:
How can we relaunch the checks? @chienandalu @pedrobaeza

pedrobaeza:
do rebase and force push

chienandalu (approved):

website_sale_coupon_reward_line_discriminant/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,9 @@
+* `Domatix <https://www.domatix.com>`_:
+
+  * Álvaro López
+
+
+* `Tecnativa <https://www.tecnativa.com>`_:
```
pedrobaeza, line 6: No need to include us, as this module is yours.

website_sale_coupon_reward_line_discriminant/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
...
+This module is useful when used in conjunction with others like 'sale_coupon_multi_gift' or 'sale_coupon_multiplier_free_product', allowing you to make offers and the customer to buy more if they want.
```
pedrobaeza, line 2: Leave an empty line in between for proper recognition of a new paragraph.

pedrobaeza (changes requested):
Very few formal questions. Apart from the comments inline, please name the commit properly:

```
[ADD] sale_coupon_reward_line_driscriminant

Enable the possibility of adding more products of a gifted product as new line
in the order.
```

PR #35, opened 2021-11-30
[FIX] Don't remove related program unless all related order lines are being removed.

alvaro-domatix:
Using sale_coupon_multi_gift module we found a bug:

If you create a sale coupon program with reward type = multigift, gifted products can have different taxes. Let's say if you have product A (with 21% tax) in your cart a coupon program gives away product B (with 21% tax) and product C (with 4% tax).

After this, we apply a discount code of 10% over all the order and it works correctly. When we want to "Pay Now", after completing the address and invoice details, the 10% discount code disappears.
This happens because "sale_coupon" module in this process creates a discount line for each tax (including the taxes of the two gifted products) and then it realizes that it is not necessary because the price is 0. So, in the process of removing this line, it removes the discount coupon program (as you are removing "the" discount line) while you still having in your cart some products (with other taxes) that can get the discount.

You can test it following my first lines. I propose this code to solve the problem. I accept suggestions and improvements (if you think the problem can be solved in any other way I'll do it.

Note: I think it's not posible to avoid the creation of the "useless" discount line because it does it Odoo and does it because the OCA module. So what I have done is not to remove the coupon program unless all the lines related to the program are being deleted.

### DIFF
--- a/sale_coupon_multi_gift/models/sale_order.py
+++ b/sale_coupon_multi_gift/models/sale_order.py
@@ -3,6 +3,10 @@
 from odoo import _, models
 from odoo.fields import first
 
+from odoo.addons.sale_management.models.sale_order import (
+    SaleOrderLine as ManagementSaleOrderLine,
+)
+
 
 class SaleOrder(models.Model):
     _inherit = "sale.order"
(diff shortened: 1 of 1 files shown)


alvaro-domatix:
I would appreciate some feedback before doing any test for the codecov @chienandalu @pedrobaeza

chienandalu (commented):
Thanks  @alvaro-domatix Yes, it's a pain to extend sale_coupon... I have to study the code as I'd prefer a simpler solution. I guess the issue has to do with the website refreshing the promotions everytime you load the controller.

PR #40, opened 2022-02-07
[13.0][FIX] sale_coupon_mass_mailing: Add groups to view.

victoralmau:
Add groups to view.

Please @CarlosRoca13 and @pedrobaeza can you review it?

@Tecnativa TT34374

### DIFF
--- a/sale_coupon_mass_mailing/views/sale_coupon_program_view.xml
+++ b/sale_coupon_mass_mailing/views/sale_coupon_program_view.xml
@@ -6,6 +6,10 @@
             name="inherit_id"
             ref="sale_coupon.sale_coupon_program_view_form_common"
         />
+        <field
+            name="groups_id"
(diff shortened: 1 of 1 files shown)


CarlosRoca13 (approved):
LGTM

sale_coupon_mass_mailing/views/sale_coupon_program_view.xml
```diff
@@ -13,6 +13,7 @@
                     type="object"
                     icon="fa-envelope"
                     name="action_mailing_count"
+                    groups="mass_mailing.group_mass_mailing_user"
```
pedrobaeza: This should be at view level, not here. If not, the error will persist.

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #41, opened 2022-02-07
[ADD] sale_coupon_reward_add_product

ivantodorovich:
When using promotions and coupons to reward a free product, the rewarded product needs to be added to the order before applying the coupon.

This module allows you to configure promotions in such a way that the rewarded product is added automatically if it's missing.

### DIFF
--- a/sale_coupon_reward_add_product/models/sale_order.py
+++ b/sale_coupon_reward_add_product/models/sale_order.py
@@ -1,0 +1,40 @@
+# Copyright 2022 Camptocamp SA (https://www.camptocamp.com).
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _is_reward_in_order_lines(self, program):
+        if not self.env.context.get("check_reward_in_order_lines", True):
+            return True
+        return super()._is_reward_in_order_lines(program)
+
+    def _prepare_reward_product_line_vals(self, program):
+        return {
+            "order_id": self.id,
+            "product_id": program.reward_product_id.id,
+            "product_uom": program.reward_product_id.uom_id.id,
+        }
+
+    def _create_reward_line(self, program):
+        # OVERRIDE to add the reward product to the order automatically
+        if program.reward_product_add_if_missing:
+            reward_product_lines = self.order_line.filtered(
+                lambda line: line.product_id == program.reward_product_id
+            )
+            order_quantity = sum(reward_product_lines.mapped("product_uom_qty"))
+            if order_quantity < program.reward_product_quantity:
+                qty_to_add = float(program.reward_product_quantity) - order_quantity
+                line_vals = self._prepare_reward_product_line_vals(program)
+                line_vals.update(product_uom_qty=qty_to_add)
+                line_vals.update(
+                    self.env["sale.order.line"].play_onchanges(
+                        line_vals, line_vals.keys()
+                    )
+                )
+                self.env["sale.order.line"].create(line_vals)
+        return super()._create_reward_line(program)
--- a/sale_coupon_reward_add_product/__init__.py
+++ b/sale_coupon_reward_add_product/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/sale_coupon_reward_add_product/__manifest__.py
+++ b/sale_coupon_reward_add_product/__manifest__.py
@@ -1,0 +1,17 @@
+# Copyright 2022 Camptocamp SA (https://www.camptocamp.com).
+# @author Julien Coux <julien.coux@camptocamp.com>
+# @author Iván Todorovich <ivan.todorovich@camptocamp.com>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Sale Coupon Automatic free product as normal",
+    "summary": "Sale Coupon Automatic free product as normal",
+    "version": "14.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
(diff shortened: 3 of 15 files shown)


sale_coupon_reward_add_product/models/sale_order.py
```diff
@@ -0,0 +1,37 @@
...
+            reward_product_lines = self.order_line.filtered(
+                lambda line: line.product_id == program.reward_product_id
+            )
+            order_quantity = sum(reward_product_lines.mapped("product_uom_qty"))
+            if order_quantity < program.reward_product_quantity:
+                qty_to_add = float(program.reward_product_quantity) - order_quantity
+                line_vals = self._prepare_reward_product_line_vals(program)
+                line = self.env["sale.order.line"].new(line_vals)
+                line.product_uom_qty = qty_to_add
+                line.product_id_change()
```
simahawk: would be better to use onchange_helper to have all onchanges played properly

simahawk (approved):
LG, just code review

damdam-s (approved):
LG thanks

sale_coupon_reward_add_product/models/sale_order.py
chienandalu: Or `odoo.tests.Form` which doesn't pull extra dependencies

ivantodorovich: So, I tested both approaches and it turns out `Form` is ~60% faster than `onchange_helper`. 
Not having to depend on an extra module is the cherry on top ❤️

rousseldenis: Many times debated. Form is not intended to be used outside tests... If onchange helper is slower than Form, I would be pleased to have metrics to fix it if needed but that is not the result we had

simahawk: @ivantodorovich could you provide some metrics? I'm curious :)

chienandalu:
By the way, an alternative approach for this would be `sale_coupon_multi_gift` or `sale_coupon_multiplier_free_product` which not only don't require the customer to manually add the rewarded product, they add the discount line with the product itself at a 100% discount. This is specially handy when sale margins are involved...

sale_coupon_reward_add_product/models/sale_order.py
ivantodorovich: > Many times debated. Form is not intended to be used outside tests

Makes sense, indeed.

> @ivantodorovich could you provide some metrics? I'm curious :)

So, I can't reach the same conclusion.
After measuring again it seems `onchange_helper` is actually faster. 🤔 
I don't remember what I did before, but it may be related to the project I was evaluating on, as I did find something curious

Take this example:

```py
class HelperTest(models.Model):
    _name = "helper.test"

    name = fields.Char()
    changed = fields.Integer()
    computed = fields.Integer(compute="_compute_computed")
    computed_group = fields.Integer(compute="_compute_computed_group", groups="base.group_portal")
    stored = fields.Integer(compute="_compute_stored", store=True)

    @api.onchange("name")
    def _onchange_name(self):
        print("_onchange_name")
        self.changed = len(self.name or "")
    
    @api.depends("name")
    def _compute_computed(self):
        print("_compute_computed")
        for rec in self:
            rec.computed = len(rec.name or "")

    @api.depends("name")
    def _compute_computed_group(self):
        print("_compute_computed_group")
        for rec in self:
            rec.computed_group = len(rec.name or "")
    
    @api.depends("name")
    def _compute_stored(self):
        print("_compute_stored")
        for rec in self:
            rec.stored = len(rec.name or "")

    @api.model
    def create_with_form(self):
        with Form(self) as form:
            form.name = "Testing"

    @api.model
    def create_with_helper(self):
        vals = {"name": "Testing"}
        vals.update(self.play_onchanges(vals, vals.keys()))
        self.create(vals)

```

```py
>>> self.env["helper.test"].create_with_form()
_onchange_name
_compute_computed
_compute_stored
_onchange_name
_compute_computed
_compute_stored
_compute_stored
_compute_computed

>>> self.env["helper.test"].create_with_helper()
_onchange_name
_compute_computed
_compute_computed_group
_compute_stored

```

- Both are running onchanges
- Both are computing non-stored computed fields (seems useless IMO, for our use case)
- Both are computing stored computed fields
- `Form` is less efficient, recomputing several times 🤔 
- `Form` is not computing fields it doesn't have access to, whilst `onchange_helper` is

Non-stored computed fields usually include fields like `_compute_helpdesk_ticket_count`, which will run sql queries to compute. Besides being useless to compute this, it may raise AccessError if the user doesn't have enough rights

So bottom line I think an improvement can be done to avoid computing non-stored fields altogether

chienandalu: I was having an issue with my `13.0` module which does a similar job to this one (see https://github.com/OCA/sale-promotion/pull/60). Finally in a real world scenario the larger the sale order was (number of lines) the worst `Form()` would perform even just initializing the form object with the record.

By the way, I'd say we both just need to get the proper price, and the rest can be prepared manually. For that, the `sale_order.onchage_product()` method would be enough.

oca_dependencies.txt
```diff
@@ -1,2 +1,3 @@
 # See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
 delivery-carrier
+server-tools
```
simahawk: we use PIP mode https://github.com/OCA/sale-promotion/blob/13.0/.copier-answers.yml#L5
You can trash this file.

simahawk: ah, funny... on v13 PIP, on v14 OCA https://github.com/OCA/sale-promotion/blob/14.0/.copier-answers.yml

can we switch it?

ivantodorovich: - https://github.com/OCA/sale-promotion/pull/61

ivantodorovich: done

PR #42, opened 2022-02-09
[13.0][FIX] sale_coupon_multi_gift: mixed rewards

chienandalu:
When we've got mixed optional rewards we need to tell for sure which reward line belongs a sale line to so we can apply the right product when the reward is updated.

cc @Tecnativa TT34266

- [x] TODO: add test

please review @pedrobaeza @CarlosRoca13

### DIFF
--- a/sale_coupon_multi_gift/__manifest__.py
+++ b/sale_coupon_multi_gift/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Coupons multi gift",
     "summary": "Allows to configure multiple gift rewards per promotion",
-    "version": "13.0.2.0.0",
+    "version": "13.0.3.0.0",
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

chienandalu:
I have to test if the options are kept without the initial context. If not, maybe I have to add an extra field to ensure it

pedrobaeza:
@chienandalu any news?

chienandalu:
Yes, I did that. Tests would be nice though...

pedrobaeza:
Did you test if the options are kept without the initial context?

chienandalu:
> Did you test if the options are kept without the initial context?

Indeed they didn't. Added tests and a field to keep the chosen option on the line.

PR #43, opened 2022-02-09
[FIX] sale_coupon_limit: improve rule times used count

chienandalu:
Sometimes Odoo doesn't remove the rules properly until the promotions
are refreshed. This could lead to count mismatches. We wan't to ensure
the count looking at the lines instead.

cc @Tecnativa TT34266

ping @pedrobaeza

### DIFF
--- a/sale_coupon_limit/models/sale_coupon_rule.py
+++ b/sale_coupon_limit/models/sale_coupon_rule.py
@@ -115,11 +115,19 @@ class SaleCouponRuleSalesmenLimit(models.Model):
                         ]
                     )
                     continue
-                salesman_limit.rule_times_used = self.env["sale.order"].search_count(
-                    [
-                        "|",
-                        ("no_code_promo_program_ids", "in", [program["id"]]),
-                        ("code_promo_program_id", "=", [program["id"]]),
-                        ("user_id", "=", salesman_limit.rule_user_id.id),
-                    ]
+                # We need to ensure that the promotion is indeed applied in the lines
+                # since a link in the sale.order can have traces of promotion links
+                # even when their lines are removed. This needs to look in the related
+                # user_id column of the sale_order table, so maybe not the most
+                # performant query. It can be improved in the future.
+                salesman_limit.rule_times_used = len(
+                    self.env["sale.order.line"].read_group(
+                        [
+                            ("coupon_program_id", "=", program["id"]),
+                            ("order_id.user_id", "=", salesman_limit.rule_user_id.id),
+                            ("state", "!=", "cancel"),
+                        ],
+                        ["order_id"],
+                        ["order_id"],
+                    )
                 )
--- a/sale_coupon_limit/__manifest__.py
+++ b/sale_coupon_limit/__manifest__.py
@@ -10,6 +10,10 @@
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "maintainers": ["chienandalu"],
     "license": "AGPL-3",
-    "depends": ["sale_coupon", "sale_commercial_partner"],
+    "depends": [
+        "sale_coupon",
+        "sale_commercial_partner",
+        "sale_coupon_order_line_link",
+    ],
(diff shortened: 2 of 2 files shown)


sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -115,11 +115,12 @@ def _compute_rule_times_used(self):
...
-                        ("code_promo_program_id", "=", [program["id"]]),
-                        ("user_id", "=", salesman_limit.rule_user_id.id),
-                    ]
+                        ("coupon_program_id", "=", [program["id"]]),
+                        ("order_id.user_id", "=", salesman_limit.rule_user_id.id),
+                        ("state", "!=", "cancel"),
+                    ],
+                    ["order_id"],
+                    ["order_id"],
                 )
```
pedrobaeza, line 133: I miss here something for counting the results (`len`?)

chienandalu: Sure :)

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -115,11 +115,14 @@ def _compute_rule_times_used(self):
...
-                    [
-                        "|",
-                        ("no_code_promo_program_ids", "in", [program["id"]]),
-                        ("code_promo_program_id", "=", [program["id"]]),
-                        ("user_id", "=", salesman_limit.rule_user_id.id),
-                    ]
+                salesman_limit.rule_times_used = len(
+                    self.env["sale.order.line"].read_group(
+                        [
+                            ("coupon_program_id", "=", [program["id"]]),
```
pedrobaeza: Why between brackets?

sale_coupon_limit/models/sale_coupon_rule.py
```diff
@@ -115,11 +115,14 @@ def _compute_rule_times_used(self):
...
+                            ("order_id.user_id", "=", salesman_limit.rule_user_id.id),
```
pedrobaeza, line 127: Isn't this field at line level?

pedrobaeza (commented):
The previous code shouldn't be changed as well?

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: A remanent. Changed. Funny that the domain was interpreted anyway

sale_coupon_limit/models/sale_coupon_rule.py
chienandalu: No. I already checked it. The field is indexed though so probabily the performance will be acceptable.

chienandalu:
> The previous code shouldn't be changed as well?

You mean the coupon part? That's ok as it is I think. Coupons are better controlled with states.

sale_coupon_limit/models/sale_coupon_rule.py
pedrobaeza: It's not a question of index, but I see that there's no `auto_join=True`, so this will be decomposed in 2 queries:

- One getting the orders with such user.
- Another for sales lines with `order_id IN ...` a a big bunch of IDs, which is not good for the performance when this list goes high.

chienandalu: We can do it in two steps: order ids for that user and then order lines.

pedrobaeza: I'm afraid the result is going to be the same. There are 2 options:

- Put auto_join=True, affecting if the are specific record rules for visibility different from header and lines.
- Put a related stored field in lines for the user.

Anyways, let's keep it for now this way, and put a comment about the performance in the line, and I fast track it.

pedrobaeza (approved):

pedrobaeza:
Fast tracking it:

/ocabot merge patch

PR #45, opened 2022-02-25
[14.0][ADD]gift_card module

Kev-Roche:
Core module for Gift Cards as payments.
depends on <https://github.com/OCA/server-tools/pull/2303>
cc @hparfr, @PierrickBrun


gift_card/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "maintainers": ["Kev-Roche"],
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "sale",
+        "account",
+        "base_generate_code",
+        "l10n_generic_coa",
```
rousseldenis: @Kev-Roche  You cannot do that

gift_card/models/account_move.py
```diff
@@ -0,0 +1,52 @@
...
+# @author Kévin Roche <kevin.roche@akretion.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    # copy from sale_usability (avoiding unwanted dependencies)
```
rousseldenis: If you want really this field here, you should change its name to something like 'sale_gift_card_ids' to avoid conflict with another module.

But IMHO, this has to be moved in a separate module

gift_card/models/account_move.py
```diff
@@ -0,0 +1,52 @@
...
+    def action_post(self):
+        res = super().action_post()
+        for move in self:
+            move._create_gift_card()
+        return res
+
+    def _create_gift_card(self):
+        for sale in self.sale_ids:
+            for line in sale.order_line:
+                tmpl = self.env["gift.card.template"].search(
```
rousseldenis: I recommend to add a gift_cart_template_ids (one2many) on product template in order to avoid a search in the loop.

gift_card/__manifest__.py
Kev-Roche: I remove it (was for test purpose)

gift_card/models/account_move.py
Kev-Roche: I deleted it, it was not very helpfull anymore.

gift_card/models/account_move.py
Kev-Roche: Good idea

luis-ron (changes requested):
Hi @Kev-Roche ,

I have made a functional review and I have detected that multicurrency is not taken into account.

Inside the gift cards there is a field "currency" which does not collect any information (both without activating the multicurrency option and once activated).

Do you plan to make an improvement? 
Do you consider the use case?

Thank you in advance.

Kev-Roche:
> Hi @Kev-Roche ,
> 
> I have made a functional review and I have detected that multicurrency is not taken into account.
> 
> Inside the gift cards there is a field "currency" which does not collect any information (both without activating the multicurrency option and once activated).
> 
> Do you plan to make an improvement? Do you consider the use case?
> 
> Thank you in advance.

Hi @luis-ron ,
Thank you for your time and for highlighting this issue, I will solve it ASAP.

gift_card/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+Core module for Gift Cards as payment method, allowing to buy gift
+cards and use them as payments.
+Gift cards can be related to a partner (beneficiary) or activated by code.
```
petrus-v, line 3: In odoo v15.0 there is a gift_card module, does this module a back port from odoo 15 ? what will be the plan on the next version for this module ?

Can we leave a message here about what's the plan for next version and how it's related ?

gift_card/models/account_move.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def action_post(self):
+        res = super().action_post()
+        for move in self:
+            move._create_gift_card()
+        return res
+
+    def _create_gift_card(self):
+        for sale in self.invoice_line_ids.mapped("sale_line_ids").mapped("order_id"):
+            for line in sale.order_line:
```
hparfr, line 19: why this link to sales ? 
we can't create giftcard directly from invoice ?

PR #47, opened 2022-04-19
[14.0][ADD] sale_promotion_domain_product_discount

ivs-cetmix:

PR #48, opened 2022-04-20
[14.0][ADD] sale_promotion_discount_in_field

GabbasovDinar:


francesco-ooops (changes requested):
if "discount in line" is selected, "Reward Line Product" should not be visible as there will be no reward line
image

francesco-ooops (changes requested):
"Discount in line" should not be allowed if settings > sales > discount flag is not active, or it can lead to unclear behavior on SO line

image

francesco-ooops:
@GabbasovDinar for the rest, LGTM!

francesco-ooops (approved):
Functional review ok!

elvise (approved):
LGTM!

francesco-ooops:
@OCA/crm-sales-marketing-maintainers can you review? thanks!

elvise:
@chienandalu can you review this PR? Thanks

elvise:
@OCA/crm-sales-marketing-maintainers can you review please ?

elvise:
@OCA/crm-sales-marketing-maintainers is there a good soul who can make a review?  :)

chienandalu (commented):
Nice addon. Thanks :) 

The main issue with this kind of extension is that we can't rely on core methods to properly count the applied promos as we're not using the automatic product created by the promo.

To achieve it you can use `sale_coupon_order_line_link` (there's a migration PR I think). For example:

https://github.com/OCA/sale-promotion/blob/13.0/sale_coupon_multiplier_free_product/models/sale_coupon_program.py#L110

francesco-ooops:
@chienandalu thanks for the feedback!
could you take a look at the migration PR? https://github.com/OCA/sale-promotion/pull/53

elvise:
@chienandalu #53 it's mandatory for your suggestion, can you check there ?

> Nice addon. Thanks :)
> 
> The main issue with this kind of extension is that we can't rely on core methods to properly count the applied promos as we're not using the automatic product created by the promo.
> 
> To achieve it you can use `sale_coupon_order_line_link` (there's a migration PR I think). For example:
> 
> https://github.com/OCA/sale-promotion/blob/13.0/sale_coupon_multiplier_free_product/models/sale_coupon_program.py#L110

sale_promotion_discount_in_field/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+{
```
simahawk, line 4: license headers missing everywhere

sale_promotion_discount_in_field/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "website": "https://github.com/OCA/sale-promotion",
+    "depends": ["sale_coupon"],
+    "external_dependencies": {},
+    "demo": [],
+    "data": [
+        "views/coupon_program_view.xml",
+    ],
+    "qweb": [],
+    "installable": True,
+    "application": False,
```
simahawk, line 20:
```suggestion
    "data": [
        "views/coupon_program_view.xml",
    ],
```

sale_promotion_discount_in_field/models/sale_order.py
```diff
@@ -0,0 +1,131 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _create_new_no_code_promo_reward_lines(self):
+        """
+        Override core method to save program with new discount apply rule
+        """
+        self.ensure_one()
+        order = self
```
simahawk: can you make this a bit more readable? Maybe you can save some indentation and split this method into smaller pieces?

sale_promotion_discount_in_field/models/sale_order.py
```diff
@@ -0,0 +1,131 @@
...
+    def _get_reward_line_values(self, program):
+        """
+        Overrider core method to set discount for line if program has
+        'discount_line' reward type
+        """
+        if program.reward_type == "discount_line":
+            self._set_reward_discount_for_lines(program)
+            return []
+        else:
+            return super()._get_reward_line_values(program)
```
simahawk: for instance
```suggestion
        if program.reward_type != "discount_line":
            return super()._get_reward_line_values(program)
        self._set_reward_discount_for_lines(program)
        return []           
```
Hmm here you are writing in a method that is supposed to "get" something?

sale_promotion_discount_in_field/models/sale_order.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def _set_reward_discount_for_lines(self, program):
+        """
+        Update discount field by program
+        """
+        discount = program.discount_percentage
+        lines = self._get_paid_order_lines()
+        if program.discount_apply_on == "cheapest_product":
```
simahawk, line 24: same here: please split this

sale_promotion_discount_in_field/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+No configuration is required to use this module
```
simahawk: you can trash this file then

sale_promotion_discount_in_field/tests/test_promotion_discount_in_field.py
```diff
@@ -0,0 +1,228 @@
+from odoo.tests import tagged
+from odoo.tests.common import SavepointCase
+
+
+@tagged("post_install", "-at_install")
+class CetmixTestSaleCouponProgram(SavepointCase):
+    def setUp(self):
```
simahawk: pls rewrite this to use `setUpClass` and set tracking_disable.

Example: https://github.com/OCA/wms/blob/14.0/shopfloor_base/tests/common.py#L99-L108

simahawk (changes requested):

chienandalu:
> @chienandalu thanks for the feedback! could you take a look at the migration PR? #53

Now is merged :)

elvise:
@simahawk what do you think now ?

GabbasovDinar:
> Nice addon. Thanks :)
> 
> The main issue with this kind of extension is that we can't rely on core methods to properly count the applied promos as we're not using the automatic product created by the promo.
> 
> To achieve it you can use `sale_coupon_order_line_link` (there's a migration PR I think). For example:
> 
> https://github.com/OCA/sale-promotion/blob/13.0/sale_coupon_multiplier_free_product/models/sale_coupon_program.py#L110

Thank you for review, can you give me more info about the  `sale_coupon_order_line_link` module. How can I use it?

ivs-cetmix:
> Nice addon. Thanks :)
> 
> The main issue with this kind of extension is that we can't rely on core methods to properly count the applied promos as we're not using the automatic product created by the promo.
> 
> To achieve it you can use `sale_coupon_order_line_link` (there's a migration PR I think). For example:
> 
> https://github.com/OCA/sale-promotion/blob/13.0/sale_coupon_multiplier_free_product/models/sale_coupon_program.py#L110

Checked this point but still cannot get it clear. Do you mean we need to inherit this method in the module you mentioned? Or this is a general approach advice?

chienandalu:
> > To achieve it you can use `sale_coupon_order_line_link` (there's a migration PR I think). For example:
> > https://github.com/OCA/sale-promotion/blob/13.0/sale_coupon_multiplier_free_product/models/sale_coupon_program.py#L110
> 
> Checked this point but still cannot get it clear. Do you mean we need to inherit this method in the module you mentioned? Or this is a general approach advice?

It'd be something like this. Anyway, thinking twice, do you set the `is_reward_line` flag in the lines with the discount? Otherwise it won't work anyway...

```python
    def _compute_order_count(self):
        """Discount in field programs need a different domain to compute the order count"""
        discount_line_programs = self.filtered(lambda x: x.reward_type == "discount_line")
        super(SaleCouponProgram, self - discount_line_programs)._compute_order_count()
        product_data = self.env["sale.order.line"].read_group(
            [
                ("coupon_program_id", "in", discount_line_programs.ids,),
                ("is_reward_line", "=", True),
            ],
            ["product_id"],
            ["product_id"],
        )
        mapped_data = {m["product_id"][0]: m["product_id_count"] for m in product_data}
        for program in discount_line_programs:
            program.order_count = mapped_data.get(program.reward_product_id.id, 0)
```

GabbasovDinar:
> > > To achieve it you can use `sale_coupon_order_line_link` (there's a migration PR I think). For example:
> > > https://github.com/OCA/sale-promotion/blob/13.0/sale_coupon_multiplier_free_product/models/sale_coupon_program.py#L110
> > 
> > 
> > Checked this point but still cannot get it clear. Do you mean we need to inherit this method in the module you mentioned? Or this is a general approach advice?
> 
> It'd be something like this. Anyway, thinking twice, do you set the `is_reward_line` flag in the lines with the discount? Otherwise it won't work anyway...
> 
> ```python
>     def _compute_order_count(self):
>         """Discount in field programs need a different domain to compute the order count"""
>         discount_line_programs = self.filtered(lambda x: x.reward_type == "discount_line")
>         super(SaleCouponProgram, self - discount_line_programs)._compute_order_count()
>         product_data = self.env["sale.order.line"].read_group(
>             [
>                 ("coupon_program_id", "in", discount_line_programs.ids,),
>                 ("is_reward_line", "=", True),
>             ],
>             ["product_id"],
>             ["product_id"],
>         )
>         mapped_data = {m["product_id"][0]: m["product_id_count"] for m in product_data}
>         for program in discount_line_programs:
>             program.order_count = mapped_data.get(program.reward_product_id.id, 0)
> ```

This is not appropriate for our case, since the discount is written in the discount field and can be applied to all order lines

What if I will use it like this (and also it's not required add to depends the sale_coupon_order_line_link module)

```python

    def _compute_order_count(self):
        discount_line_programs = self.filtered(lambda x: x.reward_type == "discount_line")
        super(CouponProgram, self - discount_line_programs)._compute_order_count()
        sale_order_obj = self.env["sale.order"]
        for program in discount_line_programs:
            program.order_count = sale_order_obj.search_count(
                [
                    "|",
                    ("no_code_promo_program_ids", "in", program.ids),
                    ("code_promo_program_id", "in", program.ids),
                ],
            )
```

ivs-cetmix:
Looks reasonable. @chienandalu what do you think?

chienandalu:
> Looks reasonable. @chienandalu what do you think?

Ok, maybe the only way to achieve it in your case. Try to test at least these cases:

- Promotion number limit.
- Update the discount.
- Remove the discount.
- Coupons.

Also try to make the search in the compute the more optimal possible.

francesco-ooops:
@chienandalu and @simahawk how does it look now?

ivs-cetmix:
@simahawk how does it look now?

francesco-ooops:
@chienandalu can you take a look?

francesco-ooops:
@simahawk can you also take a look at #47 's sister PR - this one?

Do you think it's in good shape now? Thanks :)

francesco-ooops:
@OCA/crm-sales-marketing-maintainers can you help us move this PR forward? Thanks!

sale_promotion_discount_in_field/models/coupon_program.py
```diff
@@ -0,0 +1,52 @@
...
+
+from odoo import _, api, models
+from odoo.exceptions import UserError
+
+
+class CouponProgram(models.Model):
+    _inherit = "coupon.program"
+
+    @api.onchange("reward_type")
+    def _onchange_reward_type(self):
```
chienandalu: Maybe too generic as a method name and it could collide with third modules

sale_promotion_discount_in_field/models/sale_order.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        Override core method to save program with new discount apply rule
+        """
+        self.ensure_one()
+        order = self
+        programs = order._get_applicable_no_code_promo_program()
+        programs = (
+            programs._keep_only_most_interesting_auto_applied_global_discount_program()
+        )
+        if any([program.reward_type == "discount_line" for program in programs]):
```
chienandalu: My feel is this isn't right. What happens with product and delivery rewards? Why isn't `_get_reward_values_discount` overrinding enough? There you'll have the program and you'll be able to prepare the line values accordingly.

GabbasovDinar: we cannot use the `_get_reward_values_discount` because in original method `_create_new_no_code_promo_reward_lines` expects some values to save to order_line:

`self.write({'order_line': [(0, False, value) for value in self._get_reward_line_values(program)]})`

I will try to another method

GabbasovDinar: I agree with you, I changed the method

francesco-ooops:
@chienandalu if you have a bit of time today we'd love to have your feedback :)

sale_promotion_discount_in_field/models/sale_order.py
```diff
@@ -0,0 +1,161 @@
...
+    def _get_reward_line_values(self, program):
+        """
+        Overrider core method to set discount for line if program has
+        'discount_line' reward type
+        """
+        if program.reward_type == "discount_line":
+            self._set_reward_discount_for_lines(program)
+            return []
+        else:
+            return super()._get_reward_line_values(program)
```
chienandalu:
```suggestion
        return super()._get_reward_line_values(program)
```

sale_promotion_discount_in_field/models/sale_order.py
```diff
@@ -0,0 +1,161 @@
...
+                    )
+                    .mapped("discount_line_product_id")
+                )
+                lines = lines.filtered(
+                    lambda x: x.product_id
+                    in (program.discount_specific_product_ids | free_product_lines),
+                )
+            for line in lines:
+                line.discount = discount
+        elif program.discount_apply_on == "domain_product":
```
chienandalu: With the refactor of that module this condition would be different, isn't it?

Anyway, why not having a `_get_discount_in_field_lines` method that could be overriden. That way, the only thing to do here would be something like:

```python
def _set_reward_discount_for_lines(self, program):
    if program.discount_apply_on == "cheapest_product":
        lines = self._get_cheapest_line()
    else:
        lines = self._get_discount_in_field_lines()
    lines.discount = program.discount_percentage
```

And that method would be something like:

```python
def _get_discount_in_field_lines(self):
    lines = self._get_paid_order_lines()
    if program.discount_apply_on == "specific_products":
        free_product_lines = (
            self.env["coupon.program"]
            .search(
                [
                    ("reward_type", "=", "product"),
                    (
                        "reward_product_id",
                        "in",
                        program.discount_specific_product_ids.ids,
                    ),
                ],
            )
            .mapped("discount_line_product_id")
        )
        lines = lines.filtered(
            lambda x: x.product_id
            in (program.discount_specific_product_ids | free_product_lines),
        )
     return lines
```

sale_promotion_discount_in_field/models/sale_order.py
```diff
@@ -0,0 +1,161 @@
...
+    def _create_new_no_code_promo_reward_lines(self):
+        """
+        Override core method to save program with new reward rule
+        """
+        self.ensure_one()
+        super()._create_new_no_code_promo_reward_lines()
+        order = self
+        programs = order._get_applicable_no_code_promo_discount_line_program()
+        programs = (
+            programs._keep_only_most_interesting_auto_applied_global_discount_program()
```
chienandalu: Should we add this type of promotion to the `_is_global_discount_program` rules?

chienandalu (commented):
Thanks. Looks quite good overall :). Some comments:

sale_promotion_discount_in_field/models/sale_order.py
GabbasovDinar: I think that's not necessary

sale_promotion_discount_in_field/models/sale_order.py
GabbasovDinar: thanks for this cool refactoring :)

GabbasovDinar:
@chienandalu Thanks for your review and feedback!

I updated the code

sale_promotion_discount_in_field/models/sale_order.py
chienandalu: But then two discount promotions could be applied altogether. And even if only one has effects on the lines, the others would be linked to the sales order anyway.

This is an example where I defined two discount in fields promos. The two are linked, but only the last one is applied:
image
image

It would be something like:

```python
    def _is_global_discount_program(self):
        discount_in_field_global = (
            self.promo_applicability == "on_current_order"
            and self.reward_type == "discount_line"
            and self.discount_apply_on == "on_order"
        )
        return super()._is_global_discount_program() or discount_in_field_global
```

elvise: @GabbasovDinar what do you think ?

GabbasovDinar: In your case, after overriding the _is_global_discount_program method, I can remove the unnecessary code and it will work the same way. Now with your example I have only one applied discount program

Thank you, please check :)

elvise:
@chienandalu is it good now ? :)

francesco-ooops:
@chienandalu @OCA/crm-sales-marketing-maintainers when you have time if you can review again, we'd love to see this module merged :)

francesco-ooops:
@chienandalu @simahawk @OCA/crm-sales-marketing-maintainers weekly ping

chienandalu (approved):
I've been on holidays the last weeks :smile: LGTM. Thanks!

francesco-ooops:
@simahawk if you can check again your blocking review then this would be done, thank you!

simahawk (approved):

francesco-ooops:
@simahawk @chienandalu thank you for the support!

PR #63, opened 2022-05-31
[14.0][IMP] sale_coupon_order_line_link: hook method for extra discounts

chienandalu:
If a third module adds an extra discount scope, we should:

- Allow integration returning always at least an empty object
- Allow compatibility by inheritance hooking into the method that
defines the scoped lines.

cc @Tecnativa TT36131

please review @francesco-ooops @GabbasovDinar @simahawk @elvise to unlock #47


pedrobaeza (approved):

GabbasovDinar (approved):
LGTM!

PR #68, opened 2022-06-16
[MIG] sale_coupon_criteria_multi_product: Migration to 14.0

alvaro-domatix:


sale_coupon_criteria_multi_product/models/sale_coupon_program.py
```diff
@@ -4,26 +4,26 @@
 
 
 class SaleCouponProgram(models.Model):
-    _inherit = "sale.coupon.program"
+    _inherit = "coupon.program"
 
-    sale_coupon_criteria = fields.Selection(
+    coupon_criteria = fields.Selection(
```
chienandalu: Ok with the field renames (you can also do it with the filenames using git mv for better diff). Anyway, a migration script should be made. An example:

An example: https://github.com/OCA/account-financial-tools/blob/13.0/account_move_line_sale_info/migrations/13.0.1.0.0/pre-migration.py

`openupgrade.rename_fields` reference: https://github.com/OCA/openupgradelib/blob/master/openupgradelib/openupgrade.py#L585-L611

alvaro-domatix:
Done!

francesco-ooops (approved):
functionally LGTM

chienandalu (approved):
Thanks @alvaro-domatix !

PR #69, opened 2022-06-20
[14.0] [ADD] sale_stock_coupon

paradoxxxzero:
Add `invoice_on_delivered` option to coupon/promotion programs that sets the delivered quantity of promotion on sale order line in order to invoice discount proportional to delivered quantity.


sale_stock_coupon/models/sale_order.py
```diff
@@ -0,0 +1,227 @@
+from odoo import _, api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _get_program_reward_lines(self, program):
```
chienandalu: This method name is too generic. Give it a name related to the module that uses it

sale_stock_coupon/models/stock_picking.py
```diff
@@ -0,0 +1,15 @@
...
+class Picking(models.Model):
+    _inherit = "stock.picking"
+
+    def _action_done(self):
+        # FIXME: Avoid calling recompute_coupon_lines several times on the same picking
+        rv = super()._action_done()
+        if self.sale_id._get_applied_programs_with_rewards_on_current_order().filtered(
+            lambda program: program.invoice_on_delivered
+        ):
+            self.sale_id.recompute_coupon_lines()
```
chienandalu: This could have unexpected effects. For example, a sale order has a promotion that  no longer applies when the picking is validated. It would be removed from the order! Even worse, the amount of the order would be different from the one the customer paid!

sale_stock_coupon/models/sale_order.py
```diff
@@ -0,0 +1,227 @@
...
+    @api.depends("is_reward_line")
+    def _compute_qty_delivered_method(self):
+        super(SaleOrderLine, self)._compute_qty_delivered_method()
+
+        for line in self:
+            if line.is_reward_line:
+                line.qty_delivered_method = "reward_stock_move"
+
+    @api.depends("delivered_reward_qty")
+    def _compute_qty_delivered(self):
```
chienandalu, line 215: Why not pivoting the whole logic from this method. I feel it would be much more simple. When delivered qties are triggered from the regular lines, this method goes into action an then you can:

- Check if there are promotions and if those promotions have a delivered qtys policy.
- One by one, decide how delivery lines are invoiced depending on the promotion type (general discount, cheapest product, specific products... and so on)

chienandalu (commented):
The module name is too generic. A better one could be `sale_coupon_invoice_policy` or `sale_coupon_invoice_delivered`

Some other comments below. Notice the existance of `sale_coupon_order_line_link` which could ease your life for this kind of tricky modules as well.

sale_stock_coupon/models/sale_order.py
paradoxxxzero: Good idea for a future refactor

paradoxxxzero:
Replaced with #78

PR #70, opened 2022-06-23
[ADD] sale_coupon_domain_free_product: New module 13.0

chienandalu:
This module allows to define the same reward products as the products domain. Only those products in the sale order will be rewarded.

Configuration
=============

To configure free product rewards from domains:

1. Go to *Sales > Catalog > Coupon Programs* and select or create a new one.
2. Choose the criteria of appliance you want to use and the minimum quantities, that    will be used to calculate the times the rewards can be applied.
3. On the *Reward Type* field choose *Free Product Domain*.
4. Set the proper domain and rules.
5. If you want to apply the limits for every product in the promotion limit, set the  *Strict per product limit* checkbox.

An example with the domain `[("id", "in", [1, 2])]`, strict limits and minimum quantity of 2 units:

 ===== ==================
  Qty      Products
 ===== ==================
    4   `product.product(1,)`
    2   `product.product(2,)`
    6   `product.product(3,)`
 ===== ==================

For `product.product(1,)` we'll get 2 units as reward.
For `product.product(2,)` we'll get just 1.
We'll get nothing from `product.product(3,)`

cc @Tecnativa TT37477


sale_coupon_domain_free_product/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,25 @@
...
+   *Strict per product limit* checkbox.
+
+An example with the domain `[("id", "in", [1, 2])]`, strict limits and minimum quantity
+of 2 units:
+
+
+ ===== ==================
+  Qty      Products
+ ===== ==================
+    4   `res.product(1,)`
```
pedrobaeza: res.product ?

chienandalu: Ouch :man_facepalming:

chienandalu: Fixed :)

sale_coupon_domain_free_product/models/sale_coupon_reward.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+    reward_product_max_quantity = fields.Integer(
+        string="Max reward quantity",
+        default=0,
+        help="Maximum reward product quantity (0 for no limit)",
+    )
+
+    def name_get(self):
+        """Add complete description for the multi gift reward type."""
+        res = super().name_get()
```
cesar-tecnativa: super with self - free_product_domain_promo

cesar-tecnativa (changes requested):

sale_coupon_domain_free_product/models/sale_coupon_reward.py
pedrobaeza: @chienandalu

chienandalu: Done

pedrobaeza (approved):

PR #67, opened 2022-06-15
[13.0][IMP] sale_coupon_auto_refresh: configurable triggers

chienandalu:
This can allow us to work altogether with flexible triggers like the
ones we could need with sale_coupon_criteria_order_based

cc @Tecnativa TT37479

PR #73, opened 2022-06-28
[14.0][IMP] sale_coupon_auto_refresh: configurable triggers

chienandalu:
Forward from 

- #67 

This can allow us to work altogether with flexible triggers like the ones we could need with sale_coupon_criteria_order_based

The auto-refresh in the backend is triggered over a minimum set of fields changes. If you want to extend the list of that fields:

1. Go to *Settings > Technical > Config parameters*
2. Add or update the key:

   - For `sale.order`: `sale_coupon_auto_refresh.sale_order_triggers`
   - For `sale.order.line`: `sale_coupon_auto_refresh.sale_order_line_triggers`
3. In every add the fields seperated by commas that you want to add to the recomputation
   triggers.

⚠️ After configuring or removiming a trigger a restart of Odoo is recommended so the depends are reloaded properly. Anyway it isn't mandatory and the module detects the new triggers so the auto-refresh works as expected as soon as they are set.

cc @Tecnativa TT37691

### DIFF
--- a/sale_coupon_auto_refresh/README.rst
+++ b/sale_coupon_auto_refresh/README.rst
@@ -40,6 +40,21 @@ You can set this feature on or off in every company. To do so:
 #. Go to *Sales > Configuration > Settings*
 #. In the *Pricing* section you'll find the option *Auto refresh promotions*.
 
+The auto-refresh in the backend is triggered over a minimum set of fields changes. If
+you want to extend the list of that fields:
+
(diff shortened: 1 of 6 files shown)


pedrobaeza (approved):

chienandalu:
ping @SilvioC2C

SilvioC2C (approved):
LGTM :+1:

PR #77, opened 2022-09-05
[13.0][FIX] sale_coupon_multiplier_free_product: handle invalid lines

chienandalu:
We have to put some logic redundancy here as the main method doesn't have enough granularity to avoid deleting the lines belonging to the multiplier programs when the promotions are updated. So the main module expects that the promotion lines products match with the promotion discount product (https://git.io/JWpoU) , which is not the approach in this module, where we add extra lines with the reward products themselves and the proper price tag and discount. So in this method override, we'll save those correct lines from the fire via context that the unlink method will properly catch. We also have to remove the proper invalid lines that wouldn't be detected.

cc @Tecnativa TT38806

please review @victoralmau @pedrobaeza

### DIFF
--- a/sale_coupon_multiplier_free_product/models/sale_order.py
+++ b/sale_coupon_multiplier_free_product/models/sale_order.py
@@ -85,15 +85,64 @@ class SaleOrder(models.Model):
                 lambda line: line.coupon_program_id == program and line.is_reward_line
             )
             # Remove reward line if price or qty equal to 0
-            if values.get("product_uom_qty") and values.get("price_unit"):
+            # if values.get("product_uom_qty") and values.get("price_unit"):
+            if values.get("product_uom_qty"):
                 lines.write(values)
             else:
                 lines.unlink()
 
+    def _get_applicable_programs_multiple_of(self):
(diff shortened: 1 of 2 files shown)


sale_coupon_multiplier_free_product/models/sale_order.py
```diff
@@ -90,10 +90,58 @@ def _update_existing_reward_lines(self):
...
+                lambda x: x.program_id in multiple_of_programs_to_remove
+            ).write({"state": "expired"})
+            # Detect and remove the proper unvalid program order lines
+            self.order_line.filtered(
+                lambda x: x.is_reward_line
+                and x.coupon_program_id in multiple_of_programs_to_remove
+            ).unlink()
+        # We'll catch the context in the subsequent unlink() method
+        super(
+            SaleOrder, self.with_context(valid_multiple_of_lines=valid_lines.ids)
```
pedrobaeza: You can pass recordsets instead of IDs without problem

pedrobaeza (approved):
Seems OK in the approach

victoralmau (approved):
Code review OK and very good use cases added in tests.

PR #82, opened 2022-10-28
[13.0][FIX] sale_coupon_order_line_link: missing links

chienandalu:
In some cases the `recompute_coupon_lines` method could lead to miss the previously created links. We want to ensure those links at the end of such method.

cc @Tecnativa TT40205

please review @CarlosRoca13 @victoralmau


pedrobaeza:
Shouldn't you do this on `_create_new_no_code_promo_reward_lines` method instead?

chienandalu:
@pedrobaeza done

victoralmau (approved):

pedrobaeza (approved):

PR #74, opened 2022-06-29
[ADD] sale_coupon_product_exclude: Add new module to exlude products before apply coupon

GabbasovDinar:
This module extends promotion programs allowing to apply them to orders excluding products matching the domain

PR #90, opened 2022-12-19
[14.0][FIX] sale_coupon_reward_fixed_price: Context for compatibility with sale_coupon_product_exclude

GabbasovDinar:
This module is necessary for compatibility with the sale_coupon_product_exclude module from https://github.com/OCA/sale-promotion/pull/74

### DIFF
--- a/sale_coupon_reward_fixed_price/models/sale_order.py
+++ b/sale_coupon_reward_fixed_price/models/sale_order.py
@@ -70,6 +70,9 @@ class SaleOrder(models.Model):
         """
         Update discount field by program
         """
+        # this context is used for compatibility with the sale_coupon_product_exclude module
+        self = self.with_context(current_coupon_program=program)
+
         price_unit = program.price_unit
         lines = self._get_paid_order_lines()
         if program.discount_apply_on == "cheapest_product":


francesco-ooops (approved):
@chienandalu good for you?

francesco-ooops:
@chienandalu here too :)

geomer198 (approved):
LGTM!

francesco-ooops:
@OCA/crm-sales-marketing-maintainers can we merge?

francesco-ooops:
@pedrobaeza merge?

pedrobaeza (changes requested):
Please don't be lazy on PR title (I have already changed it) and details to understand what are you doing.

For now, at least this needs a comment on the line added saying that is for compatibility with `sale_coupon_product_exclude` module. If not, someone reviewing the code will think that the context is not needed looking for it in the module code, and remove it.

But isn't any way to deduce the coupon without needing to pass such context?

DinarGabbasovArammeem:
@pedrobaeza Thank you for the review

> But isn't any way to deduce the coupon without needing to pass such context?

we cannot know the current applicable coupon since the methods that calculate the discount value do not take the current coupon as an argument. I see only three options, 1- rewrite such methods and add new arguments (this method is not safe and could potentially break many modules that use these methods), 2- add a new field that will update depending on the coupons selected (but why this complication?) and 3 - define the current context with the coupon and use it (efficient and code requires few changes).

pedrobaeza (approved):
/ocabot merge patch

PR #91, opened 2022-12-19
[14.0][IMP] sale_promotion_discount_if_field: compatibility with product exclude module

GabbasovDinar:
This fix is necessary for compatibility with the module sale_coupon_product_exclude from https://github.com/OCA/sale-promotion/pull/74

### DIFF
--- a/sale_promotion_discount_in_field/models/sale_order.py
+++ b/sale_promotion_discount_in_field/models/sale_order.py
@@ -21,6 +21,9 @@ class SaleOrder(models.Model):
         """
         Set discount for order lines
         """
+        # this context is used for compatibility with the
+        # sale_coupone_product_exclude_module
+        self = self.with_context(current_coupon_program=program)
         if program.discount_apply_on == "cheapest_product":
             lines = self._get_cheapest_line()
         else:


francesco-ooops:
@GabbasovDinar can you rebase?

francesco-ooops (approved):

francesco-ooops:
@chienandalu good for you?

francesco-ooops:
@chienandalu gentle reminder :)

geomer198 (approved):
LGTM!

francesco-ooops:
@OCA/crm-sales-marketing-maintainers can we merge?

francesco-ooops:
@pedrobaeza can we merge?

chienandalu (approved):
:+1:

chienandalu (commented):
Thinking twice... why not adding this context in the `sale_coupon_product_exclude`  `_get_reward_line_values` method?

DinarGabbasovArammeem:
> Thinking twice... why not adding this context in the `sale_coupon_product_exclude` `_get_reward_line_values` method?

in this case this method will not work if you use a module that overrides this method and "super" will not be used, for example: the module **sale_promotion_discount_in_field** overrides **_get_reward_line_values**, in which case the context will not contain this parameter

chienandalu:
Does it mean that you need to patch every module that you need to make compatible?

DinarGabbasovArammeem:
> Does it mean that you need to patch every module that you need to make compatible?

I think so

chienandalu (approved):
Ok :+1:

PR #95, opened 2023-01-18
[14.0][FIX] sale_coupon_order_line_link: missing links

chienandalu:
- Fw of #82 

In some cases the `recompute_coupon_lines` method could lead to miss the previously created links. We want to ensure those links at the end of such method.

cc @Tecnativa TT40205

please review @victoralmau @CarlosRoca13


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #94, opened 2023-01-18
[15.0][MIG] sale_coupon_order_line_link: Migration to 15.0

chienandalu:
- Standard migration

cc @Tecnativa TT37311

please review @victoralmau @CarlosRoca13

PR #76, opened 2022-08-30
[13.0][ADD] sale_coupon_multiple_code_program: New module

chienandalu:
Coupon Programs only allow one code promotion program per sale. There's no technical or practical reason to disallow it.

To use this module, you need to:

1. Go to *Sales > Master Data > Coupon Programs*
1. Configure several programs with their respective program codes.
1. Place a new order that fulfill such programs requisites.
1. Apply the programs codes.
1. You'll get every promotion applied.

cc @Tecnativa TT38413

please review @CarlosRoca13 @victoralmau

PR #96, opened 2023-01-18
[15.0][MIG] sale_coupon_limit: Migration to 15.0

chienandalu:
- [x] Partially superseeds #33 although adapting it to v15.0 to respect changes attribution
- [x] Additional adaptations
- [x] Depends on #94 
- [x] Promote to stable sale_commercial_partner https://github.com/OCA/sale-workflow/pull/2342

cc @Tecnativa TT38310

please review @pedrobaeza @victoralmau


pedrobaeza:
This must be renamed to `coupon_limit`, isn't it?

chienandalu:
> This must be renamed to `coupon_limit`, isn't it?

It depends on features from sale_coupon

pedrobaeza:
But should we split it or make it less dependant?

victoralmau (approved):
Code review OK

pedrobaeza:
This should include https://github.com/OCA/sale-promotion/pull/76/files#diff-c3be12f82d41fbacd8bf07d5ee671f0e364f35845ddb8ab6dd20d165fb7fce00

sale_coupon_limit/models/coupon_coupon.py
```diff
@@ -0,0 +1,63 @@
...
+            ("program_id", "=", self.program_id.id),
+            ("state", "=", "used"),
+        ]
+        # Customer limit rules
+        if self.program_id.rule_max_customer_application:
+            coupons_count = self.search_count(
+                domain
+                + [
+                    (
+                        "sales_order_id.commercial_partner_id",
```
pedrobaeza: The coupon may be applied to other things like a PoS order, so you should try to change this code to be more generic (isn't there a count in coupon itself for all?) and if possible, rename the module without `sale_` prefix.

chienandalu:
I don't know how I got to close this. Can you reopen?

sale_coupon_limit/models/coupon_coupon.py
chienandalu: This method is exclusive of `sales_coupon`. PoS coupons are checked in a separate js logic

pedrobaeza:
I can't open it I'm afraid

PR #105, opened 2023-02-09
[13.0][ADD] sale_coupon_commercial_partner_applicability: New module

chienandalu:
cc @Tecnativa TT41522

please review @CarlosRoca13 @sergio-teruel

PR #106, opened 2023-02-10
[13.0][ADD] sale_coupon_portal: New module to see your coupons on portal view

CarlosRoca13:
cc @Tecnativa TT41523

ping @sergio-teruel @chienandalu


sale_coupon_portal/controllers/portal.py
```diff
@@ -0,0 +1,112 @@
...
+    def _prepare_home_portal_values(self):
+        values = super()._prepare_home_portal_values()
+        coupon_count = (
+            request.env["sale.coupon"].sudo().search_count(self._get_coupons_domain())
+        )
+        values["coupon_count"] = coupon_count
+        return values
+
+    def _get_coupons_domain(self):
+        return [("partner_id", "child_of", request.env.user.commercial_partner_id.id)]
```
chienandalu: To be able to work like this, you'd need to depend on [sale_coupon_apply_commercial_partner](https://github.com/OCA/sale-promotion/pull/105) otherwise you wouldn't be able to validate the coupons that don't belong to the order's partner.

The options are:

- A bridge module
- This depending on that
- That depending on this

What do you think @sergio-teruel ?

sale_coupon_portal/templates/coupon_portal_templates.xml
```diff
@@ -0,0 +1,133 @@
...
+                </thead>
+                <tbody>
+                    <t t-foreach="coupons" t-as="coupon">
+                        <tr>
+                            <td>
+                                <a
+                                    href="#"
+                                    t-attf-onclick="navigator.clipboard.writeText(#{coupon.code});"
+                                    title="Copy to clipboard"
+                                    style="text-decoration: none;"
```
chienandalu:
```suggestion
                                    class="text-decoration-none"
```

sale_coupon_portal/templates/coupon_portal_templates.xml
```diff
@@ -0,0 +1,133 @@
...
+                        <th>Received on:</th>
+                        <th>Expires on:</th>
+                        <th />
+                    </tr>
+                </thead>
+                <tbody>
+                    <t t-foreach="coupons" t-as="coupon">
+                        <tr>
+                            <td>
+                                <a
```
chienandalu, line 67:
```suggestion
                                <!-- navigator.clipboard isn't  available on non secure environments, so this only works in https or localhost mode -->
                                <a
```

CarlosRoca13:
@chienandalu Change done and comment added

chienandalu (approved):
Thanks!

pedrobaeza:
I don't understand the dependency. It should be independent.

chienandalu:
> I don't understand the dependency. It should be independent.

That's the reason of the glue module

pedrobaeza:
Do the glue module in other PR, please, but don't block this basic one by that.

CarlosRoca13:
Done

sale_coupon_portal/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To use this module, some of your company's contacts must have a coupon.
```
pedrobaeza: Finally this one is only for the contact coupons, not for all.

pedrobaeza (approved):
/ocabot merge nobump

PR #112, opened 2023-02-17
[15.0][MIG] sale_coupon_domain_product_discount: Migration to version 15.0

stefan-tecnativa:
cc@Tecnativa TT38308


stefan-tecnativa:
This module can be splitted, i will commit it when it's done

sale_coupon_domain_product_discount/models/coupon_reward.py
```diff
@@ -35,7 +35,7 @@ def _compute_discount_specific_product_ids(self):
         Otherwise it just masks the stored field"""
         domain_product = self.env.context.get("promo_domain_product")
         if domain_product:
-            self.discount_specific_product_ids = self.env["product.product"]._search(
+            self.discount_specific_product_ids = self.env["product.product"].search(
```
chienandalu: Why?

stefan-tecnativa: Using "_search" method directly was giving me an error about wrong value of a field.  I tested in this way and worked, but I didn't investigate more the error.

chienandalu: Well, performance is crucial here so you should find out what that error might be due to.

chienandalu:
You can split it now

coupon_domain_product_discount/tests/test_domain_product_discount.py
```diff
@@ -0,0 +1,70 @@
...
+                            "applied_on": "3_global",
+                            "compute_price": "formula",
+                            "base": "list_price",
+                        },
+                    )
+                ],
+            }
+        )
+        # Let's prepare a sale order to play with
+        sale_form = Form(cls.env["sale.order"])
```
chienandalu: This declaration should go in the sales module

chienandalu (changes requested):

coupon_domain_product_discount/tests/test_domain_product_discount.py
stefan-tecnativa: Done

chienandalu (approved):
Thanks :)

pedrobaeza (commented):
The icon.png file is incorrect.

pedrobaeza (approved):
/ocabot merge nobump

PR #116, opened 2023-02-21
[13.0][ADD] sale_coupon_portal_commercial_partner_applicability: New module to see your company coupons on portal view

CarlosRoca13:
cc @Tecnativa TT41523

This PR depends on:

- [x] https://github.com/OCA/sale-promotion/pull/105
- [x] https://github.com/OCA/sale-promotion/pull/106

ping @sergio-teruel @chienandalu


chienandalu (approved):

pedrobaeza:
The names are incorrect. It should be reversed: `sale_coupon_portal_commercial_partner_apply` or `sale_coupon_portal_commercial_partner_applicability`, but I think there's something wrong in the way you are accessing the applicability, as it should be a central way to modify only one place (maybe the partner domain), for not requiring all of these glue modules.

chienandalu:
The partner domain has nothing to do with the `sale.coupon` partner field. There's no central method to hook into so the only way to ensure that we include those coupons available in the commercial entity level is to add these glue modules, I'm afraid

pedrobaeza:
Isn't there a method to say if a coupon is valid or not? Can't you hook into it? Anyway, that's independent from the module name. Please change it.

chienandalu:
> Isn't there a method to say if a coupon is valid or not? 

But only in relation to the current sales order context.

CarlosRoca13:
Name changed

PR #119, opened 2023-03-01
[13.0][FIX] sale_coupon_domain_product_discount: fix error when sending coupon by email

pilarvargas-tecnativa:
cc @Tecnativa TT41921

When the option to apply discount on specific products is checked, it expects that there are records in the field where those products are defined. If this option is checked and there are no products defined, an error occurs when generating the report that is attached to the email to send the customer their coupon because it tries to read the records of these products and they do not exist.
This has been corrected by including this option in the report to adapt it to this module and not to fail by creating a description field that will be printed in the report to send the information to the client.

@chienandalu @stefan-tecnativa  please review


sale_coupon_domain_product_discount/report/sale_coupon_report_template.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="report_coupon" inherit_id="sale_coupon.report_coupon">
+        <xpath
+            expr="//t[@t-if='len(o.program_id.discount_specific_product_ids) > 1']"
+            position="after"
+        >
+            <t t-elif="o.program_id.discount_apply_on_domain_product">
+                <t t-set="text">OFF ON SOME PRODUCTS*</t>
```
chienandalu, line 9: Think in the final result. This is how it looks right now. It doesn't make much sense. Use a similar technique to the one that core uses to print the list of products but with the description

image

pilarvargas-tecnativa: corrected, thanks!

stefan-tecnativa:
Reviewed functionality, it seems right to me

pilarvargas-tecnativa:
@stefan-tecnativa can you approve this PR if you think everything is correct?

stefan-tecnativa (approved):

sale_coupon_domain_product_discount/README.rst
```diff
@@ -14,14 +14,14 @@ Coupon Promotion Product Domain Discount
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fsale--promotion-lightgray.png?logo=github
-    :target: https://github.com/OCA/sale-promotion/tree/13.0/sale_coupon_domain_product_discount
+    :target: https://github.com/OCA/sale-promotion/tree/15.0/sale_coupon_domain_product_discount
```
pedrobaeza: This change is not correct. Please run again the README generator to go back to 13.0 definitions.

pilarvargas-tecnativa: corrected, thanks

chienandalu (approved):
Let's go with it

PR #120, opened 2023-03-02
[15.0][MIG] sale_coupon_multiplier_free_product: Migration to version 15.0

stefan-tecnativa:
cc@Tecnativa TT37310


chienandalu (changes requested):
Tests are failing...

sale_coupon_multiplier_free_product/models/sale_coupon.py
```diff
@@ -4,10 +4,13 @@
...
 class SaleCoupon(models.Model):
-    _inherit = "sale.coupon"
+    _inherit = "coupon.coupon"
 
-    def _check_coupon_code(self, order):
-        message = super()._check_coupon_code(order)
+    def _check_coupon_code(self, order_date, partner_id, **kwargs):
+        order = kwargs.get("order")
+        message = super()._check_coupon_code(
+            order.date_order.date(), order.partner_id.id, order=order
```
chienandalu: This isn't right. You should pass the args as they come not assuming the values in your override

sale_coupon_multiplier_free_product/models/sale_coupon_program.py
```diff
@@ -75,9 +75,12 @@ def _filter_not_ordered_reward_programs(self, order):
         return programs
 
     def _is_valid_product(self, product):
+        """This method was removed, it's functionallity
```
chienandalu: Then is useless to override it

sale_coupon_multiplier_free_product/models/sale_order.py
```diff
@@ -141,9 +142,10 @@ def _remove_invalid_reward_lines(self):
                 and x.coupon_program_id in multiple_of_programs_to_remove
             ).unlink()
         # We'll catch the context in the subsequent unlink() method
-        super(
+        res = super(
             SaleOrder, self.with_context(valid_multiple_of_lines=valid_lines)
         )._remove_invalid_reward_lines()
+        return res
```
chienandalu: You don't need to make an assignment if you're going to return that value immediately after it
```suggestion
        return super(
            SaleOrder, self.with_context(valid_multiple_of_lines=valid_lines)
        )._remove_invalid_reward_lines()
```

chienandalu (changes requested):

sale_coupon_multiplier_free_product/models/sale_order.py
stefan-tecnativa: All changes done, thank you for all the suggestions.

sergio-teruel:
ping @chienandalu

chienandalu (approved):
:+1:

sergio-teruel (approved):

PR #88, opened 2022-12-13
[MIG] sale_coupon_criteria_multi_product: Migration to 15.0

carolina-domatix:

PR #122, opened 2023-03-03
[15.0][MIG] sale_coupon_criteria_multi_product: Migration to version 15.0

stefan-tecnativa:
cc @Tecnativa TT39032

This PR superseeds #88 


pedrobaeza:
There's a merge commit, so the history is not correct. You should redo it.

coupon_criteria_multi_product/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,9 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Pedro M. Baeza
+  * David Vidal
+  * Stefan Ungureanu
+
+* `Domatix <https://www.domatix.com>`_:
+
+  * Álvaro López Oró
```
chienandalu, line 9: Let's add @carolina-domatix as contributor

sale_coupon_criteria_multi_product/__manifest__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Coupons multi product criteria",
```
chienandalu:
```suggestion
    "name": "Coupons multi product criteria in sales",
```

chienandalu (changes requested):
Thanks @stefan-tecnativa Some comments below:

- Check tests :red_circle: 
- Reference to the original PR we're superseeding here in the PR description
- Fixup this commit:

image

coupon_criteria_multi_product/readme/CONTRIBUTORS.rst
stefan-tecnativa: Done

sale_coupon_criteria_multi_product/__manifest__.py
stefan-tecnativa: Done, thanks for all the suggestions.

chienandalu (changes requested):
Great :) Check the tests anyway :)

sale_coupon_criteria_multi_product/__manifest__.py
```diff
@@ -10,6 +10,5 @@
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "maintainers": ["chienandalu"],
     "license": "AGPL-3",
-    "depends": ["sale_coupon"],
-    "data": ["views/sale_coupon_program_views.xml", "security/ir.model.access.csv"],
+    "depends": ["coupon_criteria_multi_product"],
```
chienandalu: It should depend on `sale_coupon` as well, don't you think?
```suggestion
    "depends": ["coupon_criteria_multi_product", "sale_coupon"],
```

sale_coupon_criteria_multi_product/models/coupon_program.py
```diff
@@ -1,30 +1,11 @@
...
-        inverse_name="program_id",
-    )
-
-    @api.onchange("coupon_criteria")
-    def _onchange_coupon_criteria(self):
-        """Clear domain so we clear some other fields from the view"""
-        if self.coupon_criteria == "multi_product":
-            self.rule_products_domain = False
+class CouponProgram(models.Model):
+    _inherit = ["coupon.program", "coupon.criteria"]
```
chienandalu: Why? :open_mouth: (BTW, this looks like the real reason behind your previous test errors)

coupon_criteria_multi_product/models/coupon_criteria.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class CouponCriteria(models.Model):
+    _name = "coupon.criteria"
```
chienandalu, line 8: You should do a migration script to rename the model. One simple example: https://github.com/OCA/stock-logistics-warehouse/blob/13.0/scrap_reason_code/migrations/13.0.1.1.0/pre-migration.py#L14

chienandalu (changes requested):

coupon_criteria_multi_product/models/coupon_criteria.py
stefan-tecnativa: Would this need to be done in all similar cases?

coupon_criteria_multi_product/migrations/13.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 Tecnativa - Stefan Ungureanu
```
chienandalu: This script will never be executed. Can you figure out why?

sale_coupon_criteria_multi_product/models/coupon_program.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models
+
+
+class CouponProgram(models.Model):
+    _inherit = "coupon.program"
+    _name = "coupon.program"
```
chienandalu: Still wrong...
```suggestion
```

chienandalu (changes requested):

coupon_criteria_multi_product/migrations/13.0.1.0.0/pre-migration.py
stefan-tecnativa: Was it a problem with the folder name?

chienandalu: Indeed :) When the module is updated we have to match the proper module version and that should be greater than the version we're migrating from

coupon_criteria_multi_product/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "name": "Coupons multi product criteria",
+    "summary": "Allows to set as promotion criteria multi-product conditions",
+    "version": "15.0.1.0.0",
+    "development_status": "Production/Stable",
+    "category": "Sale",
+    "website": "https://github.com/OCA/sale-promotion",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "maintainers": ["chienandalu"],
+    "license": "AGPL-3",
+    "depends": ["sale_coupon"],
```
chienandalu: It doesn't make sense to split the module if we don't change the dependecy chain, don't you think?
```suggestion
    "depends": ["coupon"],
```

coupon_criteria_multi_product/models/coupon_criteria.py
chienandalu: If the model is created in your module and you change its name in a later iteration, you must do so. Otherwise you'll lose your previous data.

chienandalu (changes requested):

coupon_criteria_multi_product/__manifest__.py
stefan-tecnativa: Totally in agreement, what a cluelessness

coupon_criteria_multi_product/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,13 @@
...
+from openupgradelib import openupgrade
+
+_model_renames = [
+    ("sale.coupon.criteria", "coupon.criteria"),
+]
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_models(env.cr, _model_renames)
```
pedrobaeza, line 14: Missing `rename_tables` method.

pedrobaeza (changes requested):

pedrobaeza (approved):

chienandalu (approved):
LGTM :+1:

PR #124, opened 2023-03-28
[15.0][FIX] coupon_domain_product_discount: fix error when sending coupon by email

pilarvargas-tecnativa:
Fix done in v13: https://github.com/OCA/sale-promotion/pull/119/files

cc @Tecnativa TT41937

@stefan-tecnativa @chienandalu can you review, please? :)

When the option to apply discount on specific products is checked, it expects that there are records in the field where those products are defined. If this option is checked and there are no products defined, an error occurs when generating the report that is attached to the email to send the customer their coupon because it tries to read the records of these products and they do not exist. This has been corrected by including this option in the report to adapt it to this module and not to fail by creating a description field that will be printed in the report to send the information to the client.


coupon_domain_product_discount/README.rst
```diff
@@ -14,14 +14,14 @@ Coupon Promotion Product Domain Discount
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fsale--promotion-lightgray.png?logo=github
-    :target: https://github.com/OCA/sale-promotion/tree/15.0/coupon_domain_product_discount
+    :target: https://github.com/OCA/sale-promotion/tree/13.0/coupon_domain_product_discount
```
pedrobaeza: Regenerate the README to avoid this and similar changes

pilarvargas-tecnativa: thanks!

chienandalu (approved):
:+1:

stefan-tecnativa (approved):
Looks good to me!

PR #127, opened 2023-04-20
[15.0][OU-ADD] coupon_multi_gift: Add migration script

chienandalu:
cc @Tecnativa TT39005

please review @pedrobaeza


pedrobaeza:

PR #128, opened 2023-04-20
[15.0][OU-ADD] coupon_criteria_multi_product: migration script

chienandalu:
Coming from a previous version we expect to have sale_coupon_criteria_multi_product installed despite the split

cc @Tecnativa TT39032

review @pedrobaeza @stefan-tecnativa

### DIFF
--- a/coupon_criteria_multi_product/migrations/15.0.1.0.0/pre-migration.py
+++ b/coupon_criteria_multi_product/migrations/15.0.1.0.0/pre-migration.py
@@ -9,7 +9,18 @@ _model_renames = [
 _table_renames = [("sale_coupon_criteria", "coupon_criteria")]
 
 
+def install_new_modules(env):
+    openupgrade.logged_query(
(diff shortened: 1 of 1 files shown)


coupon_criteria_multi_product/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -9,7 +9,18 @@
 _table_renames = [("sale_coupon_criteria", "coupon_criteria")]
 
 
+def install_new_modules(env):
```
sergio-teruel, line 12: Private method?? _install_new_modules...

chienandalu: What for? Private methods only work at RPC level

stefan-tecnativa (approved):
Tested, look good to me.

pedrobaeza:

PR #129, opened 2023-04-20
[15.0][OU-ADD] coupon_incompatibility: Migration script

chienandalu:
cc @Tecnativa TT38309

please review @pedrobaeza @stefan-tecnativa


stefan-tecnativa (approved):
Tested, looks great.

pedrobaeza:

PR #132, opened 2023-05-11
[15.0][ADD] sale_coupon_validity_notification: New module

pilarvargas-tecnativa:
This module is designed to notify that a coupon has expired in case it is being applied to a sales order.

cc @Tecnativa TT42509

@chienandalu @pedrobaeza please review :)


sale_coupon_validity_notification/models/coupon.py
```diff
@@ -0,0 +1,36 @@
...
+                        "Coupon <a href=# data-oe-model=coupon.coupon data-oe-id=%(id)s>"
+                        "%(coupon)s</a> has expired.",
+                        id=coupon.program_id,
+                        coupon=coupon.program_id.name,
+                    )
+                )
+                order.activity_schedule(
+                    "mail.mail_activity_data_warning",
+                    summary=_("Coupon expired"),
+                    user_id=order.user_id.id or order.partner_id.user_id.id,
```
chienandalu: Do you think is a good idea putting an activity to the order's partner? :sweat_smile:

pilarvargas-tecnativa: Of course not! :woman_facepalming: 
corrected :smiley:

sale_coupon_validity_notification/models/coupon.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class CouponCoupon(models.Model):
+    _inherit = "coupon.coupon"
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("state", False) == "expired":
+            for coupon in self.filtered(
+                lambda r: r.state == "expired" and r.sales_order_id
```
chienandalu: This should be done **before** the super call so we can check the previous state of the coupons and notify only on those that were valid and now are getting expired. You aren't filtering only generated coupons either.

sale_coupon_validity_notification/models/coupon.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, models
+
+
+class CouponCoupon(models.Model):
+    _inherit = "coupon.coupon"
+
+    def write(self, vals):
+        if vals.get("state", False) == "expired":
```
chienandalu: If there's no state in the keys this comparison is going to fail
```suggestion
        if vals.get("state", False) == "expired":
```

pedrobaeza: False > "" ?

pilarvargas-tecnativa: thanks :)

sale_coupon_validity_notification/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,3 @@
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+  * Pilar Vargas
```
pedrobaeza, line 4: Put David here as the designer contributor.

pilarvargas-tecnativa:
@chienandalu Can you check if the translations are correct, please?

chienandalu (approved):

CarlosRoca13 (approved):

PR #133, opened 2023-05-11
[13.0][ADD] sale_coupon_financial_risk: New module

stefan-tecnativa:
cc @Tecnativa TT43281

This module avoids sending a coupon by mail when the costumer has exceeded it's financial risk restriction.

@chienandalu please review!


sale_coupon_financial_risk/models/sale_order.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_confirm(self):
+        exception_msg = self.evaluate_risk_message(
+            self.partner_invoice_id.commercial_partner_id
+        )
+        if exception_msg:
```
chienandalu: If the risk is bypassed we should send the mail (test this as well)
```suggestion
        if exception_msg and not self.env.context.get("bypass_risk"):
```

chienandalu (commented):
Well done :+1: 

Just squash your commits and check pre-commit issues

chienandalu (commented):
Squash both commits into a single one

stefan-tecnativa:
> Squash both commits into a single one

Done, I was checking the pre-commit when you commented.

sale_coupon_financial_risk/models/sale_order.py
```diff
@@ -0,0 +1,22 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def action_confirm(self):
+        exception_msg = self.evaluate_risk_message(
+            self.partner_invoice_id.commercial_partner_id
+        )
+        if exception_msg and not self.env.context.get("bypass_risk"):
+            self = self.with_context(risk_exceeded=True)
+        elif self._context.get("risk_exceeded"):
```
chienandalu:
```suggestion
        elif self.env.context.get("risk_exceeded"):
```

chienandalu (approved):

chienandalu:
please review @carlosdauden

carlosdauden (approved):

PR #149, opened 2023-08-29
[16.0][MIG] website_sale_coupon_page > website_sale_loyalty_page: Migration to version 16.0

pilarvargas-tecnativa:
Renamed to website_sale_loyalty_page + Migration to 16.0

cc @Tecnativa TT44378

@CarlosRoca13 @chienandalu please review


chienandalu (approved):
:+1:

CarlosRoca13 (approved):

pedrobaeza:
Shouldn't this be renamed to `website_loyalty_page` or `website_sale_loyalty_page`?

pilarvargas-tecnativa:
> Shouldn't this be renamed to `website_loyalty_page` or `website_sale_loyalty_page`?

Depends on website_sale_loyalty so I think website_sale_loyalty_page is correct.

pilarvargas-tecnativa:
> > Shouldn't this be renamed to `website_loyalty_page` or `website_sale_loyalty_page`?
> 
> Depends on website_sale_loyalty so I think website_sale_loyalty_page is correct.

But now that I check the dependencies I think it's more correct that it depends on loyalty

pilarvargas-tecnativa:
> > > Shouldn't this be renamed to `website_loyalty_page` or `website_sale_loyalty_page`?
> > 
> > 
> > Depends on website_sale_loyalty so I think website_sale_loyalty_page is correct.
> 
> But now that I check the dependencies I think it's more correct that it depends on loyalty

I'm not sure, I'm making a mess with the changes that have been made in odoo, this is to show the promotions in the ecommerce, so it should depend on website_sale and loyalty and the bridge module is website_sale_loyalty, is it correct? In this case it would be fine renamed

pedrobaeza:
You have to check if the tools to discriminate the "coupons" to show are in `loyalty` or in `website_sale_loyalty`, and decide accordingly.

chienandalu:
Strictly it could be that this module only depended on loyalty and website, but it the practice that information is only usable on the eCommerce, so...

pedrobaeza:
A page can be rendered without website, so maybe the `website` dependency is not needed as well.

chienandalu:
Well, the website.published.mixin is being used so that dep can't be dropped

pilarvargas-tecnativa:
After analysing the dependencies, they should be loyalty and website. But now comes the dilemma of website_sale_loyalty, does it make sense for a site without e-commerce to show the promotions? I don't know if there will be any case in which, for example, without e-commerce they use the website as a presentation or catalogue and in that case it would be interesting to have this option available.

pedrobaeza:
If technically not needed, let's not put it for not forcing such installation. On mostly cases, they will install on their own, but not because we need it.

pilarvargas-tecnativa:
I autocorrect myself:

The correct dependency is website_sale_loyalty, only the promotions published and associated to the website are shown, this is set by the "website_id" field of the mixin "website.multi.mixin" which is defined for the "loyalty.program" model in the website_sale_loyalty module. So the dependency is correct and it is well renamed.

ping @pedrobaeza

PR #151, opened 2023-09-18
[16.0][MIG] sale_coupon_order_line_link: Migration to version 16.0

pilarvargas-tecnativa:
cc @Tecnativa TT44349

TODO: Renamed to sale_loyalty_order_line_link and script for migration

Please @chienandalu can you check it? If it is correct I rename it


sale_coupon_order_line_link/models/sale_order.py
```diff
@@ -114,29 +93,15 @@ def _link_reward_lines(self, programs):
...
-        res = super()._create_new_no_code_promo_reward_lines()
-        for order in self:
-            order._link_reward_generated_lines(order.order_line.coupon_program_id)
-        return res
-
 
 class SaleOrderLine(models.Model):
     _inherit = "sale.order.line"
 
-    coupon_program_id = fields.Many2one(
```
chienandalu: Why removing this?

sale_coupon_order_line_link/reports/sale_report.py
```diff
@@ -7,22 +7,17 @@
 class SaleReport(models.Model):
     _inherit = "sale.report"
 
-    coupon_program_id = fields.Many2one(
```
chienandalu: The same. We're still interested in this info.

sale_coupon_order_line_link/views/sale_order_views.xml
```diff
@@ -5,46 +5,44 @@
         <field name="inherit_id" ref="sale.view_order_form" />
         <field name="arch" type="xml">
             <xpath expr="//page[@name='other_information']" position="inside">
-                <group
-                    attrs="{'invisible': [('no_code_promo_program_ids', '=', []), ('code_promo_program_id', '=', False), ('applied_coupon_ids', '=', []), ('generated_coupon_ids', '=', [])]}"
-                >
+                <group attrs="{'invisible': [('applied_coupon_ids', '=', [])]}">
                     <group name="promo_info" string="Promotions">
-                        <field
+                        <!-- <field
```
chienandalu: :thinking:

chienandalu (changes requested):
Yes, the rename would be correct IMO.

Some remarks:

sale_coupon_order_line_link/models/sale_order.py
pilarvargas-tecnativa: Because there is already a field called reward_id in the sale.order.line model that contains the same information. Would it be necessary to duplicate it?

sale_coupon_order_line_link/views/sale_order_views.xml
pilarvargas-tecnativa: This view needs to be revised because the fields have changed and may need to be added to the sale.order model to display them.

sale_coupon_order_line_link/models/sale_order.py
chienandalu: But it's not the same record. You can get the data from the reward_id converting it to a related field, but we don't want to miss this information. So:

- Rename the field to loyalty_program_id
- Make it related stored to reward_id

sale_coupon_order_line_link/views/sale_order_views.xml
pilarvargas-tecnativa: These fields no longer exist in the sale.order model nor have they been renamed, I would have to add them to the model to get the promotion/coupon applied and in the case that a coupon is generated also add a field to display those generated coupons. I don't know if this is necessary to complete the functionality of the module or if this would be a parallel functionality.

sale_coupon_order_line_link/views/sale_order_views.xml
```diff
@@ -5,46 +5,21 @@
...
-                            widget="many2many_tags"
-                        />
                     </group>
                 </group>
             </xpath>
             <xpath
                 expr="//field[@name='order_line']/form//field[@name='tax_id']"
                 position="after"
             >
-                <field name="coupon_program_id" readonly="True" />
```
chienandalu: This should remain

sale_coupon_order_line_link/views/sale_order_views.xml
```diff
@@ -61,7 +36,6 @@
                 expr="//field[@name='order_line']/tree/field[@name='price_unit']"
                 position="after"
             >
-                <field name="coupon_program_id" readonly="True" optional="hide" />
```
chienandalu: And this

sale_coupon_order_line_link/views/sale_order_views.xml
```diff
@@ -88,7 +62,6 @@
         <field name="inherit_id" ref="sale.view_order_line_tree" />
         <field name="arch" type="xml">
             <field name="name" position="after">
-                <field name="coupon_program_id" readonly="True" optional="hide" />
```
chienandalu: And this

sale_coupon_order_line_link/models/sale_order.py
```diff
@@ -115,28 +98,20 @@ def _link_reward_generated_lines(self, programs):
...
-        comodel_name="coupon.program",
-        ondelete="restrict",
-        string="Coupon Program",
+    loyalty_program_id = fields.Many2one(
```
chienandalu: I'm missing the migration script for this change

chienandalu (changes requested):

sale_coupon_order_line_link/views/sale_order_views.xml
pilarvargas-tecnativa: But if the reward_id field it is relating to already exists in the view, wouldn't that be showing duplicate information?
image

sale_coupon_order_line_link/models/sale_order.py
```diff
@@ -115,28 +98,20 @@ def _link_reward_generated_lines(self, programs):
...
+        related="reward_id", store=True, string="Loyalty Program"
```
chienandalu:
```suggestion
        related="reward_id.program_id", store=True, string="Loyalty Program"
```

sale_coupon_order_line_link/views/sale_order_views.xml
chienandalu: It isn't the same (check my other comment because your related is wrong) Where's Odoo adding that field to the view? I couldn't see it

chienandalu (changes requested):

sale_coupon_order_line_link/models/sale_order.py
pilarvargas-tecnativa: Done!

sale_coupon_order_line_link/views/sale_order_views.xml
pilarvargas-tecnativa: No way, I had a dirty database.

sale_coupon_order_line_link/models/sale_order.py
pilarvargas-tecnativa: Corrected the relation and the view fields.

chienandalu (commented):
Add the migration script

sale_coupon_order_line_link/models/sale_order.py
pilarvargas-tecnativa: Script done. If you give me the OK I will proceed to renaming

sale_coupon_order_line_link/migrations/16.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def _fill_reward_id_field(env):
+    for line in env["sale.order.line"].filtered("loyalty_program_id"):
+        line.reward_id = env["reward.program"].search(
+            [["program_id", "=", line.loyalty_program_id.id]]
+        )
```
chienandalu: Better do this in pre and with sql for performance and to avoid to trigger ORM operations

chienandalu (changes requested):

sale_coupon_order_line_link/migrations/16.0.1.0.0/post-migration.py
pilarvargas-tecnativa: I doubted whether such a field will already exist in pre

chienandalu: It probably won't. You should try to pre-create the column: 

```sql
        ALTER TABLE sale_order_line
        ADD COLUMN IF NOT EXISTS reward_id INTEGER;
```

pilarvargas-tecnativa:
Continued in https://github.com/OCA/sale-promotion/pull/152

PR #152, opened 2023-09-21
[16.0][MIG] sale_coupon_order_line_link > sale_loyalty_order_line_link: Migration to version 16.0

pilarvargas-tecnativa:
Renamed to sale_loyalty_order_line_link + Migration to 16.0

cc @Tecnativa TT44349

ping @chienandalu @stefan-tecnativa


sale_loyalty_order_line_link/migrations/16.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+def _fill_reward_id_field(env):
+    for line in env["sale.order.line"].filtered("loyalty_program_id"):
+        line.reward_id = env["loyalty.reward"].search(
```
pilarvargas-tecnativa: @chienandalu 
If the reward_id field does not exist in pre, then neither does the loyalty_reward table.
The SQL query would look like this (it could also be executed in post):
```
UPDATE sale_order_line AS sol
SET reward_id = (
    SELECT lr.id
    FROM loyalty_reward AS lr
    WHERE sol.loyalty_program_id = lr.program_id
)
WHERE sol.loyalty_program_id IS NOT NULL AND sol.loyalty_program_id != '';
```
Unless the base migration respects the ids of the reward_id field in coupon.program (which is not clear to me) and then after creating the field it would be executed:
```
UPDATE sale_order_line AS sol
SET reward_id = (
    SELECT cp.id
    FROM coupon_program AS cp
    WHERE sol.loyalty_program_id = cp.id
)
WHERE sol.loyalty_program_id IS NOT NULL AND sol.loyalty_program_id != '';
```

stefan-tecnativa (approved):
Functionally tested. LGTM 👍️

sale_loyalty_order_line_link/migrations/16.0.1.0.0/post-migration.py
chienandalu: Once you get to your migration script every should be settled down in the base modules, so put yourself in an scenario where everything is properly renamed and transformed. You just have tu pull the loyalty_reward ids that match the loyalty_program ids you already have. Is that simple.

pilarvargas-tecnativa: OK, done, thanks!

sale_loyalty_order_line_link/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,31 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, _renamed_fields)
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS reward_id INTEGER;
+        UPDATE sale_order_line AS sol
```
chienandalu, line 28: Do a statement per `logged_query`

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -46,13 +26,16 @@ def _get_discounted_lines(self, program):
...
-            return self._get_cheapest_line()
-        elif program.discount_apply_on == "specific_products":
+        elif program.discount_applicability == "cheapest":
+            return self._cheapest_line()
+        elif program.discount_applicability == "specific":
             return self.order_line.filtered(
-                lambda x: x.product_id in program.discount_specific_product_ids
+                lambda x: x.product_id
+                in self.env["product.product"].search(
+                    program._get_discount_product_domain()
```
chienandalu: Why don't you use the `_get_specific_discountable_lines` method (https://github.com/odoo/odoo/blob/adae60fc532cf6a853815f840cb49e93eb008cb7/addons/sale_loyalty/models/sale_order.py#L194)

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -7,36 +7,16 @@
...
         the proper ids to write them into the related lines"""
         res = super().write(vals)
-        reward_lines = [
-            x[2]
-            for x in vals.get("order_line", [])
-            if len(x) > 2 and x[0] == 0 and x[2].get("is_reward_line")
-        ]
+        reward_lines = self.order_line.filtered(
+            lambda x: x.loyalty_program_id and x.is_reward_line
+        )
```
chienandalu, line 15: Not using the ORM is deliberate here for performance reasons. This change isn't needed.

**Anyway**, I'm thinking that now that we have the base module setting the reward for us maybe we could simplify the logic triggering the _link_reward_lines as a compute that depends on `reward_id`.

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -176,9 +151,9 @@ def _filter_related_program_lines(self, program):
         the product criteria rules, we can extend this method to return the proper
         records."""
         # No domain means that any product meets the promotion rules
-        if not program.rule_products_domain:
-            return self
-        domain = safe_eval(program.rule_products_domain)
+        if not program.program_id.rule_ids.product_domain:
```
chienandalu: This got me thinking for a while :sweat_smile: as the nomenclature is really misleading. If you're going to use rewards use rewards, but don't name it otherwise.

And I'd say you shouldn't as you're doing an unnecessary roundtrip

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -176,9 +151,9 @@ def _filter_related_program_lines(self, program):
...
+            return self.filtered(lambda x: not x.is_reward_line)
+        domain = safe_eval(program.program_id.rule_ids.product_domain)
```
chienandalu: This is going to crash for sure if there are several rules. But you've got a more reliable way to obtain the domain so you can combine them: https://github.com/odoo/odoo/blob/adae60fc532cf6a853815f840cb49e93eb008cb7/addons/loyalty/models/loyalty_rule.py#L117-L128

sale_loyalty_order_line_link/tests/test_sale_loyalty_order_line_link.py
```diff
@@ -71,14 +86,20 @@ def setUpClass(cls):
             line_form.tax_id.clear()
             line_form.tax_id.add(cls.tax_2)
         cls.sale = sale_form.save()
+        cls.sale._update_programs_and_rewards()
+        cls.wizard = (
```
chienandalu, line 90: If you're forcing the method in the previous line, why do you need to do it again in the wizard?

chienandalu (changes requested):

sale_loyalty_order_line_link/tests/test_sale_loyalty_order_line_link.py
pilarvargas-tecnativa: If there are multiple programs that are applicable to the order, this method will evaluate each program and update rewards and points accordingly. Likewise in case a program has been applied and modifications are made to the sales order.

In the case that there are multiple programs applicable, the wizard will select which program to apply.

sale_loyalty_order_line_link/models/sale_order.py
pilarvargas-tecnativa: I was looking for the method to filter the lines by specific products and didn't realise that there was already a method that returned the lines for that type of reward. Thank you very much :)

sale_loyalty_order_line_link/tests/test_sale_loyalty_order_line_link.py
chienandalu: But do you need it for the test?

pilarvargas-tecnativa: When debugging locally, the promotion is not applied automatically because it detects that there are other programs that meet the conditions to be applied, this is the standard demo data. If I delete the wizard in my database it fails because it does not apply the promotion automatically when it detects others that meet the conditions to apply. I can try to run the promotion without the wizard and test outside my database to see what happens. ....

chienandalu: Ok, don't worry about it right now. It's ok if it makes the test more reliable

sale_loyalty_order_line_link/models/sale_order.py
pilarvargas-tecnativa: I've gone through the whole code, correcting nomenclatures and o2m field accesses to avoid errors and one of the changes is applied in this part.

sale_loyalty_order_line_link/models/sale_order.py
pilarvargas-tecnativa: You're right it wasn't properly named or correct

sale_loyalty_order_line_link/models/sale_order.py
pilarvargas-tecnativa: 'is_reward_line' doesn't come as a key in the write vals of sale.order as it did in v15, that's why I made the change.

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -175,11 +149,15 @@ def _filter_related_program_lines(self, program):
...
-        domain = safe_eval(program.rule_products_domain)
-        products = self.mapped("product_id").filtered_domain(domain)
-        return self.filtered(
-            lambda x: x.product_id in products and not x.is_reward_line
-        )
+        programs_line = self.env["sale.order.line"]
+        for rule in program.rule_ids:
+            valid_product_domain = rule._get_valid_product_domain()
+            if valid_product_domain:
+                domain = safe_eval(str(valid_product_domain))
```
chienandalu: You don't need to do this at all. You're converting a python expression to string and then to a python expression again.

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -175,11 +149,15 @@ def _filter_related_program_lines(self, program):
...
-        # No domain means that any product meets the promotion rules
-        if not program.rule_products_domain:
-            return self
-        domain = safe_eval(program.rule_products_domain)
-        products = self.mapped("product_id").filtered_domain(domain)
-        return self.filtered(
-            lambda x: x.product_id in products and not x.is_reward_line
-        )
+        programs_line = self.env["sale.order.line"]
+        for rule in program.rule_ids:
```
chienandalu: Don't iterate, but append the different rules domain and make the filter at once

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -175,11 +149,15 @@ def _filter_related_program_lines(self, program):
...
+            valid_product_domain = rule._get_valid_product_domain()
+            if valid_product_domain:
+                domain = safe_eval(str(valid_product_domain))
+                products = self.mapped("product_id").filtered_domain(domain)
```
chienandalu:
```suggestion
                products = self.product_id.filtered_domain(domain)
```

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -175,11 +149,15 @@ def _filter_related_program_lines(self, program):
...
+                programs_line |= self.filtered(
+                    lambda x: x.product_id in products and not x.is_reward_line
+                )
+            else:
+                programs_line |= self.filtered(lambda x: not x.is_reward_line)
+        return programs_line
```
chienandalu: The logic was quite simple before and you don't need to make it this complex at all. Just use the previous logic taking the domain from the provided method

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -81,18 +61,18 @@ def _link_reward_discount_lines(self, program):
                 {"reward_line_ids": [(4, rl.id) for rl in tax_reward_lines]}
             )
 
-    def _link_reward_product_lines(self, program):
+    def _link_reward_product_lines(self, reward):
         """We want to link to the reward those lines that generated it as well as
         the rewarded product, that could be not in the original domain"""
         reward_lines = self.order_line.filtered(
-            lambda x: x.coupon_program_id == program
+            lambda x: x.loyalty_program_id == reward.program_id
```
chienandalu: It's useless to change this signature if what we need is the program itself. I didn't ask for it

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -46,19 +28,17 @@ def _get_discounted_lines(self, program):
...
+        elif program.discount_applicability == "cheapest":
+            return self._cheapest_line()
+        elif program.discount_applicability == "specific":
+            return self._get_specific_discountable_lines(program)
         # An extra discount scope could not depend no this module. For integration
         # purposes, allways return at least an empty object
         return self.env["sale.order.line"]
 
-    def _link_reward_discount_lines(self, program):
+    def _link_reward_discount_lines(self, reward):
```
chienandalu: The same here

sale_loyalty_order_line_link/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line AS sol
+        SET reward_id = (
+            SELECT lr.id
+            FROM loyalty_reward AS lr
+            WHERE sol.loyalty_program_id = lr.program_id
+        )
+        WHERE sol.loyalty_program_id IS NOT NULL AND sol.loyalty_program_id != '';
```
chienandalu:
```suggestion
        WHERE sol.loyalty_program_id IS NOT NULL
```

sale_loyalty_order_line_link/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.rename_fields(env, _renamed_fields)
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE sale_order_line
+        ADD COLUMN IF NOT EXISTS reward_id INTEGER;
```
chienandalu, line 22: Don't put semicolons in the statements
```suggestion
        ADD COLUMN IF NOT EXISTS reward_id INTEGER
```

sale_loyalty_order_line_link/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line AS sol
+        SET reward_id = (
+            SELECT lr.id
+            FROM loyalty_reward AS lr
+            WHERE sol.loyalty_program_id = lr.program_id
+        )
```
chienandalu: Maybe better
```suggestion
        SET sol.reward_id = lr.id
        FROM loyalty_reward lr
        WHERE sol.loyalty_program_id = lr.program_id
```

chienandalu (changes requested):

sale_loyalty_order_line_link/models/sale_order.py
chienandalu: Then let's convert these fields into computed store and trigger them from `reward_id`

pilarvargas-tecnativa: You mean convert loyalty_program_id and is_reward_line to computed?

chienandalu: No, the links between lines

chienandalu (approved):
Let's go :)

sale_loyalty_order_line_link/migrations/16.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE sale_order_line AS sol
+        SET sol.reward_id = lr.id
+        FROM loyalty_reward AS lr
+        WHERE sol.loyalty_program_id = lr.program_id
+        AND sol.loyalty_program_id IS NOT NULL;
```
pedrobaeza: No need of this condition, as you are already matching with loyalty_reward and only rows with `loyalty_program_id` filled will match.

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -1,41 +1,22 @@
...
-        ]
+        reward_lines = self.order_line.filtered(
+            lambda x: x.loyalty_program_id and x.is_reward_line
+        )
         if not reward_lines:
             return res
-        programs = self.env["coupon.program"].browse(
-            list({x.get("coupon_program_id") for x in reward_lines})
+        programs = self.env["loyalty.program"].browse(
+            list({x.loyalty_program_id.id for x in reward_lines})
```
pedrobaeza: Don't rebrowse here, but just populate the recordset:

```suggestion
        programs = reward_lines.loyalty_program_id
```

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -85,58 +66,55 @@ def _link_reward_product_lines(self, program):
...
     _inherit = "sale.order.line"
 
-    coupon_program_id = fields.Many2one(
-        comodel_name="coupon.program",
-        ondelete="restrict",
-        string="Coupon Program",
+    loyalty_program_id = fields.Many2one(
+        related="reward_id.program_id", store=True, string="Loyalty Program"
     )
+
```
pedrobaeza: Don't add empty line here

pilarvargas-tecnativa: Changes done, thanks :)

PR #155, opened 2023-09-27
[16.0][MIG] coupon_limit > loyalty_limit: Migration to version 16.0

pilarvargas-tecnativa:
Renamed to loyalty_limit

cc @Tecnativa TT44319

@chienandalu @ernesto-garcia-tecnativa please review

PR #156, opened 2023-10-02
[16.0][MIG] sale_coupon_limit > sale_loyalty_limit: Migration to version 16.0

pilarvargas-tecnativa:
Renamed to sale_loyalty_limit

Pending from:
- [x] https://github.com/OCA/sale-promotion/pull/155

cc @Tecnativa TT44347


sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,132 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _program_check_compute_points(self, programs):
+        res = super()._program_check_compute_points(programs)
+        programs_with_additional_errors = {}
+        # Iterate through the programs that initially have no errors
+        for program, result in res.items():
+            if "error" not in result:
```
chienandalu: Avoid nesting if not necessary
```suggestion
            if result == "error":
                continue
```

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,132 @@
+from odoo import _, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _program_check_compute_points(self, programs):
+        res = super()._program_check_compute_points(programs)
+        programs_with_additional_errors = {}
```
chienandalu: You already have a dict coming from `super()` just write over that one when a program can't be applied

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,132 @@
...
+        # Iterate through the programs that initially have no errors
+        for program, result in res.items():
+            if "error" not in result:
+                for rule in program.rule_ids:
```
chienandalu: Take the logic in a separate method and check here if all the rules are limited. Only in that case we flag the program in the dictionary

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,132 @@
...
+                            ),
+                        ]
+                        order_count = self.env["sale.order"].search_count(
+                            customer_domain
+                        )
+                        limit_reached = (
+                            order_count >= rule.rule_max_customer_application
+                        )
+                        if limit_reached:
+                            programs_with_additional_errors[program] = {
```
chienandalu: This isn't right. If one rule passes the check we shouldn't flag the promo

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,132 @@
...
+                            }
+                    # Salesmen limit rules
+                    if rule.rule_salesmen_limit_ids:
+                        salesman_rule = rule.rule_salesmen_limit_ids.filtered(
+                            lambda x: self.user_id == x.rule_user_id
+                        )
+                        max_rule = salesman_rule.rule_max_salesman_application
+                        times_used = salesman_rule.rule_times_used
+                        if times_used and times_used >= max_rule:
+                            programs_with_additional_errors[program] = {
```
chienandalu: Same here

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,132 @@
...
+        res = super().__try_apply_program(program, coupon, status)
+        domain = [
+            (
+                "order_line.loyalty_program_id",
+                "=",
+                program.id,
+            ),  # Buscar órdenes con líneas que tienen loyalty_program_id igual a program.id
+            ("id", "!=", self.id),  # Excluir la orden actual (self)
+        ]
+        for rule in program.rule_ids:
```
chienandalu: Same comment as above. And these two methods could share the logic

sale_loyalty_limit/models/coupon_rule.py
```diff
@@ -19,42 +19,70 @@ def _compute_rule_salesmen_limit_count(self):
...
             for salesman_limit in salesmen_limits:
                 if program in coupon_programs and not program["coupon_ids"]:
                     continue
                 elif program in coupon_programs:
+                    required_points = (
+                        self.env["loyalty.program"]
+                        .browse(program["id"])
+                        .reward_ids.mapped("required_points")
+                    )
+                    min_required_points = min(required_points) if required_points else 1
```
chienandalu: Why do we need to check this?

chienandalu (changes requested):

sale_loyalty_limit/models/coupon_rule.py
pilarvargas-tecnativa: It's a mistake of mine that I didn't correct, defining a promotion without rewards gave me a failure in the tests but it doesn't make sense for a promotion to have no rewards.

sale_loyalty_limit/models/sale_order.py
pilarvargas-tecnativa: This generates me a bit of doubt, let's say we have 2 rules, we would have two cases:
1. one of the rules has "rule_max_customer_application" set and the other one doesn't, in that case no error should be generated and the promotion is applied. (In this case the functionality of another module to enforce all the rules would be the one that would generate the error)
2. Both rules have "rule_max_customer_application" set and in this case an error must be generated to not allow the promotion to be applied (in this case is where I am looking for confirmation). If one rule has the limit set to 1 and another rule has the limit set to 2, does the lower limit prevail?

chienandalu: > If one rule has the limit set to 1 and another rule has the limit set to 2, does the lower limit prevail?

It doesn't matter. You check them individually. If one passes, the promotion is valid and you can forget about the rest

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __try_apply_program(self, program, coupon, status):
+        res = super().__try_apply_program(program, coupon, status)
+        if all(rule.rule_max_customer_application for rule in program.rule_ids):
+            error = self._check_customer_limit(program, coupon)
+            if error:
+                return error
+        # Salesmen limit rules
+        if all(rule.rule_salesmen_limit_ids for rule in program.rule_ids):
+            error = self._check_salesmen_limit(program)
```
chienandalu: Checking the core method, you don't need to do this here

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _program_check_compute_points(self, programs):
+        res = super()._program_check_compute_points(programs)
+        # Iterate through the programs that initially have no errors
+        for program, result in res.items():
+            if result == "error":
+                continue
+            # Customer limit rules
+            if all(rule.rule_max_customer_application for rule in program.rule_ids):
+                error = self._check_customer_limit(program)
```
chienandalu: This isn't the logic we need. If only one rule in the program is valid, the program is valid and we break the loop. The same with the customer limit

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,103 @@
+from odoo import _, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _check_customer_limit(self, program, coupon=False):
```
chienandalu: Just check the rule

chienandalu (changes requested):

pilarvargas-tecnativa:
@chienandalu I have gone round the whole code reducing what was not necessary:

- When calculating how many times it has been used it is not necessary to take into account the coupons, you simply have to take into account the order lines where the program that complies with the rules has been applied.

- It is only necessary to make the check in _program_check_compute_points.

Keeping in mind that "if one rule passes, they all pass":

- For the limit of uses for a client, if all rules have that limit set, it will be taken into account that if in a rule has not yet reached the limit the promotion will still apply, i.e. the highest limit set in the rules. If it is not stated in all the rules then there will be no limit to apply.

- For the limit of uses set for a seller, it will only apply in case all rules have it set. It will be taken into account whether or not the limit of the rule with the highest limit has been reached, because if one passes, they all pass, the minimum limits will be exceeded by the maximum limits.

- The strict limit rule will also be applied only if it has been established in all the rules defined in the promotion.

I hope everything is well explained and understood :)

pilarvargas-tecnativa:
What I am left wondering is that I cannot use the computed field to get the number of times a promotion has been used due to the execution of the _update_programs_and_rewards method when confirming a sales order because it does not exclude the current order.

sale_loyalty_limit/tests/test_sale_loyalty_limit.py
```diff
@@ -18,241 +28,312 @@ def _create_sale(self, partner, salesman=False):
...
-        self._apply_coupon(sale_1, "TEST-SALE-COUPON-LIMIT")
+        self._apply_promo_code(sale_1, "TEST-SALE-COUPON-LIMIT")
         self.assertTrue(bool(sale_1.order_line.filtered("is_reward_line")))
+        sale_1.action_confirm()
         # We apply it twice for partner 1...
         sale_2 = self._create_sale(self.partner_1)
-        self._apply_coupon(sale_2, "TEST-SALE-COUPON-LIMIT")
+        self._apply_promo_code(sale_2, "TEST-SALE-COUPON-LIMIT")
         self.assertTrue(bool(sale_2.order_line.filtered("is_reward_line")))
+        sale_2.action_confirm()
```
chienandalu: Why do we need to confirm it now?

sale_loyalty_limit/tests/test_sale_loyalty_limit.py
```diff
@@ -18,241 +28,312 @@ def _create_sale(self, partner, salesman=False):
...
+                            "rule_user_id": self.salesman_2.id,
+                            "rule_max_salesman_application": 2,
+                        },
+                    ),
+                ]
+            }
+        )
+        self.loyalty_promotion.reward_ids.write(
+            {"discount_product_ids": self.product_a}
+        )
```
chienandalu: Why don't you make simple assignments

chienandalu:
> What I am left wondering is that I cannot use the computed field to get the number of times a promotion has been used due to the execution of the _update_programs_and_rewards method when confirming a sales order because it does not exclude the current order.

Why would you exclude the current order?

chienandalu (commented):
> When calculating how many times it has been used it is not necessary to take into account the coupons, you simply have to take into account the order lines where the program that complies with the rules has been applied.

What about generated coupons?

> Keeping in mind that "if one rule passes, they all pass"...

I'm wondering if it makes much sense to have the limits at that level and not at loyalty.program globaly

pilarvargas-tecnativa:
> Why would you exclude the current order?

Because when confirming the sale order (action_confirm) it calls again _update_programs_and_rewards. If we set the limit to 1 use, when the promotion is applied it calls _update_programs_and_rewards and if it has not been used it will calculate 0 uses, when confirming the sales order it calls _update_programs_and_rewards again, it recalculates the uses and will take into account the use in the current sales order and eliminates the reward line, so if the limit is set to 1 it will never be applied after confirming the order. You would have to set the limit to 2 in order for it to be applied 1 time.

pilarvargas-tecnativa:
> > When calculating how many times it has been used it is not necessary to take into account the coupons, you simply have to take into account the order lines where the program that complies with the rules has been applied.
> 
> What about generated coupons?

Example of a promotion that generates coupons with a limit of 1 use for admin, 1 coupon generated by this promotion has already been used in a sales order and in the images that I show I am about to apply another generated coupon, in this case I should not allow it since the limit has been set to 1 use and it has already been used. 
generatedcoupons

pilarvargas-tecnativa:
> I'm wondering if it makes much sense to have the limits at that level and not at loyalty.program globaly

It doesn't have to, the logical thing to do is to configure all the application parameters of a promotion in the same rule, as it was done in v15, although the logic adapts to what odoo proposes.

chienandalu:
> It doesn't have to, the logical thing to do is to configure all the application parameters of a promotion in the same rule, as it was done in v15, although the logic adapts to what odoo proposes.

But I can't figure out a case where you could take advantage of multiple rules with different limits. What the purpose would be?

It could fit in a multiple rewards design, as you could stablish different limits to different rewards. But in any case I wouldn't complicate this more than necessary and simply put the limits at program level where they were

sale_loyalty_limit/tests/test_sale_loyalty_limit.py
pilarvargas-tecnativa: In this case it is not necessary, it is only necessary to generate valid coupons in promotions that generate coupons for the next sales order because if the order is not confirmed it generates the coupon but it will not be valid until it is confirmed. I have made the change in the tests so that they are only confirmed in these cases.

sale_loyalty_limit/models/sale_order.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _check_all_max_customer_application(self, program):
+        """
+        Check if all rules associated with a loyalty program have the
+        'max_customer_application' property set to True.
+
+        :param program: Loyalty program object
+        :return: True if all rules meet the condition, False otherwise
+        """
+        return bool(all(rule.max_customer_application for rule in program.rule_ids))
```
chienandalu: `all` already returns a Boolan value, and you can use mapped instead of a for. Anyway... do we need these hook methods for these simple one-liners?
```suggestion
        return all(program.rule_ids.mapped("max_customer_application"))
```

pilarvargas-tecnativa: They are not necessary when making the change at the program level, I had forgotten to delete them.

chienandalu (commented):
Tests are failing

pilarvargas-tecnativa:
> Tests are failing

Yes, I have replicated it locally, when the module is installed alone there are no failures, I am investigating which module is causing the error and why.

pilarvargas-tecnativa:
The error is being caused by sale_loyalty_partner_applicability

pedrobaeza:
Please rebase.

/ocabot migration sale_coupon_limit

chienandalu (approved):
Thanks :)

pilarvargas-tecnativa:
@CarlosRoca13 take a look please :)

CarlosRoca13 (approved):

PR #158, opened 2023-10-04
[16.0][FIX] sale_loyalty_order_line_link: Fix to avoid recursion error and test improvements

pilarvargas-tecnativa:
cc @Tecnativa TT44349

@chienandalu @stefan-tecnativa please review

### DIFF
--- a/sale_loyalty_order_line_link/models/sale_order.py
+++ b/sale_loyalty_order_line_link/models/sale_order.py
@@ -140,10 +140,14 @@ class SaleOrderLine(models.Model):
     def write(self, vals):
         """When the reward line is update we should refresh the line links as well"""
         res = super().write(vals)
-        if self.filtered(lambda x: x.is_reward_line and x.loyalty_program_id):
-            program = self.env["loyalty.program"].browse([self.loyalty_program_id.id])
+        if self.filtered(
+            lambda x: x.is_reward_line
+            and x.loyalty_program_id
+            and not x.reward_line_ids
+        ):
+            programs = self.env["loyalty.program"].browse(self.loyalty_program_id.ids)
             for order in self.mapped("order_id"):
-                order._link_reward_lines(program)
+                order._link_reward_lines(programs)
         return res
 
     def _filter_related_program_lines(self, program):
--- a/sale_loyalty_order_line_link/tests/test_sale_loyalty_order_line_link.py
+++ b/sale_loyalty_order_line_link/tests/test_sale_loyalty_order_line_link.py
@@ -164,32 +164,27 @@ class TestSaleCouponCriteriaMultiProduct(TransactionCase):
         genereate the reward line are linked to that
         """
         # Let's set up the program for product rewards
-        self.loyalty_program.reward_ids.reward_type = "product"
-        self.loyalty_program.reward_ids.reward_product_id = self.product_c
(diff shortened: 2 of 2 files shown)


sale_loyalty_order_line_link/tests/test_sale_loyalty_order_line_link.py
```diff
@@ -138,7 +138,7 @@ def test_01_coupon_order_line_link_discount(self):
     def test_02_coupon_order_line_link_discount_cheapest(self):
         """Change the program discount type to a cheapest product. Now only the chepest
         line will get the reward."""
-        self.loyalty_program.reward_ids.discount_applicability = "cheapest"
+        self.loyalty_program.reward_ids.write({"discount_applicability": "cheapest"})
```
chienandalu: This change is unnecessary

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -140,7 +140,11 @@ class SaleOrderLine(models.Model):
     def write(self, vals):
         """When the reward line is update we should refresh the line links as well"""
         res = super().write(vals)
-        if self.filtered(lambda x: x.is_reward_line and x.loyalty_program_id):
+        if self.filtered(
+            lambda x: x.is_reward_line
+            and x.loyalty_program_id
+            and not x.reward_line_ids
```
chienandalu, line 146: What if the lines change?

pilarvargas-tecnativa: If the lines change, the applicable promotions are updated, in case there is any applicable promotion a new line is generated that fulfils the conditions of the if and the old one is deleted, in case they do not fulfil the conditions of any promotion all will be deleted. Moreover, if the applicable promotions have not been manually updated when confirming, the order confirmation will do it. Is there any use case that I'm missing and may be giving an error?

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -140,10 +140,16 @@ class SaleOrderLine(models.Model):
...
         res = super().write(vals)
-        if self.filtered(lambda x: x.is_reward_line and x.loyalty_program_id):
-            program = self.env["loyalty.program"].browse([self.loyalty_program_id.id])
+        if self.filtered(
+            lambda x: x.is_reward_line
+            and x.loyalty_program_id
+            and not x.reward_line_ids
+        ):
+            programs = self.env["loyalty.program"].search(
+                [("id", "=", self.loyalty_program_id.ids)]
```
chienandalu, line 149: Careful with this
```suggestion
                [("id", "=", self.loyalty_program_id.id)]
```

sale_loyalty_order_line_link/models/sale_order.py
```diff
@@ -140,10 +140,16 @@ class SaleOrderLine(models.Model):
...
+            )
```
chienandalu:
```suggestion
            programs = self.env["loyalty.program"].browse(self.loyalty_program_id.ids)
```

pilarvargas-tecnativa: thanks!

stefan-tecnativa (approved):
Tested in runbot. The error isn't anymore.

chienandalu (approved):
:+1:

PR #162, opened 2023-10-05
[16.0][MIG] coupon_multi_gift: Migration to version 16.0

ernesto-garcia-tecnativa:
@tecnativa TT44321


ernesto-garcia-tecnativa:
@chienandalu please review

loyalty_multi_gift/views/coupon_program_views.xml
```diff
@@ -1,32 +1,14 @@
...
-                                required="1"
-                            />
-                        </tree>
-                    </field>
-                </group>
+            <xpath
+                expr="//form/sheet/group/group/field[@name='available_on']"
+                position="after"
+            >
+                <field name="get_all_rewards" />
```
chienandalu: For this, I'd add an special program type and use the method that sets the default values for that case.

chienandalu (changes requested):
Thanks @ernesto-garcia-tecnativa 

- Rename the files as well (as tipped in your loyalty_incompatibility migration)
- Let's avoid that a user could mix reward types. We should only be able to set those that add products.

loyalty_multi_gift/models/loyalty_reward.py
```diff
@@ -3,36 +3,39 @@
...
 
-class CouponReward(models.Model):
-    _inherit = "coupon.reward"
+class LoyaltyProgram(models.Model):
+    _inherit = "loyalty.program"
 
-    coupon_multi_gift_ids = fields.One2many(
-        comodel_name="coupon.reward.product_line",
-        inverse_name="reward_id",
+    loyalty_multi_gift_ids = fields.One2many(
```
chienandalu, line 9: This will require a migration script

loyalty_multi_gift/models/loyalty_reward.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, api, fields, models
+
+
+class LoyaltyProgram(models.Model):
+    _inherit = "loyalty.program"
```
chienandalu: This should be linked to the `loyalty.reward` model

loyalty_multi_gift/models/loyalty_reward.py
```diff
@@ -0,0 +1,69 @@
...
+        string="Quantity",
+        help="Reward product quantity",
+    )
+    reward_default_product_id = fields.Many2one(
+        comodel_name="product.product",
+        compute="_compute_reward_default_product_id",
+        inverse="_inverse_reward_default_product_id",
+        readonly=False,
+    )
+    reward_product_ids = fields.Many2many(
```
chienandalu, line 49: Odoo already can make gifts optional, although the scheme is different from ours. Here, for every line we could give the customer the option to choose which product to reward while Odoo offers the option of which reward (as a whole) to choose. It would be very interesting to converge although I'm not sure if we can make it happen (maybe we find limitations as we did in the multi criteria module...)

chienandalu (commented):
Could we use multi

PR #164, opened 2023-10-10
[16.0][ADD] sale_loyalty_partner_applicability: New module

pilarvargas-tecnativa:
Pending from:
- [x] https://github.com/OCA/sale-promotion/pull/171

cc @Tecnativa TT45413


sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+        return is_valid
+
+    def _program_check_compute_points(self, programs):
+        res = super()._program_check_compute_points(programs)
+        # Iterate through the programs that initially have no errors
+        for program, result in res.items():
+            if result == "error":
+                continue
+            if self._is_valid_partner(program):
+                break
```
chienandalu: Why breaking the loop? We should check every valid program in the result dict

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+            if not self._is_valid_partner(program):
```
chienandalu, line 59: Don't check it twice. It's a binary condition: you already know the result the first time you check it

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,42 @@
...
+                res[program] = {
+                    "error": _("The customer doesn't have access to this reward.")
+                }
+        return res
+
+    def __try_apply_program(self, program, coupon, status):
+        res = super().__try_apply_program(program, coupon, status)
+        if not self._is_valid_partner(program):
+            return {"error": _("The customer doesn't have access to this reward.")}
```
chienandalu, line 76: Isn't it enough to check it in `_program_check_compute_points`

pilarvargas-tecnativa: Automatically applied programs do not pass through the _program_check_compute_points method.
https://github.com/odoo/odoo/blob/16.0/addons/sale_loyalty/models/sale_order.py#L775-L779

https://github.com/odoo/odoo/blob/16.0/addons/sale_loyalty/models/sale_order.py#L652-L656

sale_loyalty_partner_applicability/models/sale_order.py
pilarvargas-tecnativa: I got confused with the rules route, sorry.

sale_loyalty_partner_applicability/models/sale_order.py
pilarvargas-tecnativa: I have reanalysed it, it is not necessary, it is redundant.

chienandalu: :smile: :+1:

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _is_valid_partner(self, program):
+        is_valid = False
+        for rule in program.rule_ids:
+            if not is_valid and rule.rule_partners_domain != "[]":
+                domain = ast.literal_eval(rule.rule_partners_domain) + [
+                    ("id", "=", self.partner_id.id)
```
chienandalu: Add a hook method to get the partner domain (just return this one for the moment) so we can use it for the migration of `coupon_commercial_partner_applicability`

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _is_valid_partner(self, program):
+        is_valid = False
+        for rule in program.rule_ids:
+            if not is_valid and rule.rule_partners_domain != "[]":
+                domain = ast.literal_eval(rule.rule_partners_domain) + [
```
chienandalu: Use `expression.AND` to concatenate botho domains

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+from odoo import _, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _is_valid_partner(self, program):
+        is_valid = False
+        for rule in program.rule_ids:
+            if not is_valid and rule.rule_partners_domain != "[]":
```
chienandalu: You don't need to check `is_valid` here, just break whenever is `True`

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+                domain = ast.literal_eval(rule.rule_partners_domain) + [
+                    ("id", "=", self.partner_id.id)
+                ]
+                is_valid = bool(self.env["res.partner"].search_count(domain))
+            else:
+                is_valid = True
```
chienandalu:
```suggestion
            if is_valid:
                break
```

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import ast
+
+from odoo import _, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _is_valid_partner(self, program):
```
chienandalu, line 29: Add a docstring with the rationale

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,34 @@
...
+                ]
+                is_valid = bool(self.env["res.partner"].search_count(domain))
+            else:
+                is_valid = True
+                break
+        return is_valid
+
+    def _program_check_compute_points(self, programs):
+        res = super()._program_check_compute_points(programs)
+        # Iterate through the programs that initially have no errors
```
chienandalu: Put it as docstring and explain the why we do this, no what we're doing

sale_loyalty_partner_applicability/models/sale_order.py
pilarvargas-tecnativa: I have put it in another, simpler way in my opinion

sale_loyalty_partner_applicability/models/sale_order.py
pilarvargas-tecnativa: done! How do you see it? :S :P

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,57 @@
...
+        Args:
+            program (recordset): The loyalty program for which partner eligibility is checked.
+        Returns:
+            bool: True if the partner is eligible for the program, False otherwise.
+        """
+        for rule in program.rule_ids:
+            partner_domain = self._get_partner_domain(rule)
+            if partner_domain != "[]":
+                domain = expression.AND(
+                    [partner_domain, [("id", "=", self.partner_id.id)]]
```
chienandalu: Well, precisely the part we're interested in is the one that's still encapsulated :sweat_smile:

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,59 @@
...
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    def _get_partner_domain(self, rule, partner_id):
+        # For now, let's return the rule's partner domain.
+        if rule.rule_partners_domain != "[]":
+            return expression.AND(
+                [
+                    ast.literal_eval(rule.rule_partners_domain),
+                    [("id", "=", partner_id.id)],
```
chienandalu: I'm think that we could get rid of the modules `*_coupon_commercial_partner_applicability` adding a simple setting here (or in the base one being purists) to check just the partner or the commercial partner children. With that we would save three modules including the glue `coupon_commercial_partner_applicability`.

pilarvargas-tecnativa: I have added the functionality of the coupon_commercial_partner_applicability module by adding an option in settings in the sales section. In addition, I have extended the module description and usage as well as added tests to check this added functionality.

pilarvargas-tecnativa: As a clarification, it is necessary to make the check in _try_apply_code for when a code is applied, because if it is done later, if it is a coupon, the code will already be applied and even if there is an error it will be at promotion level and not at coupon level.

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -0,0 +1,80 @@
...
+        Args:
+            program (recordset): The loyalty program for which partner eligibility is checked.
+        Returns:
+            bool: True if the partner is eligible for the program, False otherwise.
+        """
+        for rule in program.rule_ids:
+            partner_domain = self._get_partner_domain(rule, self.partner_id)
+            if partner_domain:
+                if self.env["res.partner"].search_count(partner_domain):
+                    return True
```
chienandalu:
```suggestion
            if partner_domain and self.env["res.partner"].search_count(partner_domain):
                return True
```

chienandalu (commented):
Looks good overall, but check CIs which are failing on the build

sale_loyalty_partner_applicability/views/res_config_settings.xml
```diff
@@ -0,0 +1,29 @@
...
+        <field name="inherit_id" ref="sale.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@id='coupon_settings']" position="after">
+                <div
+                    class="col-12 col-lg-6 o_setting_box"
+                    id="coupon_sharing"
+                    title="Allows a coupon generated for one partner to be used by any other member of your commercial entity."
+                >
+                    <div class="o_setting_left_pane">
+                        <field name="allow_coupon_sharing" />
```
ernesto-garcia-tecnativa, line 17: this fields are not added in res.config.settings

sale_loyalty_partner_applicability/models/sale_order.py
ernesto-garcia-tecnativa: maybe is unnecessary, because partner_domain get an empty list, and the search_count function return a 0

ernesto-garcia-tecnativa (changes requested):

sale_loyalty_partner_applicability/views/res_config_settings.xml
pilarvargas-tecnativa: It is added in the base module pending approval
https://github.com/OCA/sale-promotion/pull/171

sale_loyalty_partner_applicability/models/sale_order.py
pilarvargas-tecnativa: If partner_domain = [] (False) and it is not checked then self.env["res.partner"].search_count(partner_domain) will return the number of all records, this search only needs to be done in case partner_domain != [] (True)

pilarvargas-tecnativa:
> Looks good overall, but check CIs which are failing on the build

They are failing because of the PR where I add the configuration option in the base module which is not yet merged. What I don't know is why it doesn't build runboat, I have rechecked test-requirements but I don't see what could be wrong.

#171

test-requeriments-txt
```diff
@@ -0,0 +1 @@
+odoo-addon-sale_loyalty_partner_applicability @ git+https://github.com/OCA/sale-promotion.git@refs/pull/171/head#subdirectory=setup/sale_loyalty_partner_applicability
```
chienandalu:
```suggestion
odoo-addon-loyalty_partner_applicability @ git+https://github.com/OCA/sale-promotion.git@refs/pull/171/head#subdirectory=setup/loyalty_partner_applicability
```

pedrobaeza:
Please check CI

sale_loyalty_partner_applicability/tests/test_sale_loyalty_partner_applicability.py
```diff
@@ -0,0 +1,233 @@
...
+            raise ValidationError(_("No reward to claim with this coupon"))
+        coupons = self.env["loyalty.card"]
+        rewards = self.env["loyalty.reward"]
+        for coupon, coupon_rewards in status.items():
+            coupons |= coupon
+            rewards |= coupon_rewards
+        if len(coupons) == 1 and len(rewards) == 1:
+            status = order._apply_program_reward(rewards, coupons)
+            if "error" in status:
+                raise ValidationError(status["error"])
```
chienandalu: Is this condition ever tested?

sale_loyalty_partner_applicability/tests/test_sale_loyalty_partner_applicability.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+        wizard.action_apply()
+
+    def _apply_promo_code(self, order, code, no_reward_fail=True):
+        status = order._try_apply_code(code)
+        if "error" in status:
+            raise ValidationError(status["error"])
+        if not status and no_reward_fail:
+            # Can happen if global discount got filtered out in `_get_claimable_rewards`
+            raise ValidationError(_("No reward to claim with this coupon"))
```
chienandalu: Is this condition ever tested?

chienandalu (commented):
Looks good :)

Just a little doubt seeing that coverage dicrease on the tests themselves (always strange):

sale_loyalty_partner_applicability/tests/test_sale_loyalty_partner_applicability.py
pilarvargas-tecnativa: In the case of the tests of this module it is only necessary to check if there is an error when applying the code, therefore I have removed the conditions that are not necessary, thank thanks.

chienandalu (approved):
Ok, let's go :smile: :+1: 

please @CarlosRoca13 would you add your review?

chienandalu:
Reping @CarlosRoca13

CarlosRoca13 (approved):

PR #170, opened 2023-10-31
[16.0][MIG] coupon_commercial_partner_applicability: Migration to version 16.0

pilarvargas-tecnativa:
Pending from:
- [ ] https://github.com/OCA/sale-promotion/pull/164

cc @Tecnativa TT45744


pedrobaeza:
Why not integrating this into the main module?

coupon_commercial_partner_applicability/migrations/pre-migration.py
```diff
@@ -0,0 +1,10 @@
+from openupgradelib import openupgrade
+
+namespec = [
+    ("coupon_commercial_partner_applicability", "sale_loyalty_partner_applicability")
```
chienandalu, line 4: This should be in OU apriori, isn't it?

pilarvargas-tecnativa:
It closes because the migration script should not be in the module itself.

PR #171, opened 2023-10-31
[16.0][IMP] loyalty_partner_applicability: add configuration option

pilarvargas-tecnativa:
The option to allow sharing of coupons between members of the same trading entity in the settings must be in the base module.

cc @tecnativa TT45413

### DIFF
--- a/loyalty_partner_applicability/models/res_config_settings.py
+++ b/loyalty_partner_applicability/models/res_config_settings.py
@@ -1,0 +1,17 @@
+# Copyright 2023 Tecnativa - Pilar Vargas Moreno
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+
+    _inherit = "res.config.settings"
+
+    # Allows a coupon generated for one partner to be used by any other member
+    # of your commercial entity. There is a view for this option in the module
+    # sale_loyalty_partner_applicability
+    allow_coupon_sharing = fields.Boolean(
+        string="Allow coupon sharing",
+        config_parameter="allow_coupon_sharing",
+    )


ernesto-garcia-tecnativa (approved):
Code review ok

loyalty_partner_applicability/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+
+    _inherit = "res.config.settings"
+
+    allow_coupon_sharing = fields.Boolean(
+        string="Allow coupon sharing", config_parameter="allow_coupon_sharing"
```
chienandalu: Add a help string describing the effect of this: if checked, all the commercial entity members of the partner will fullfil the rule... (or something along the lines)

chienandalu (commented):
Won't you add the proper settings views?

loyalty_partner_applicability/models/res_config_settings.py
pilarvargas-tecnativa: Yes, the field is added in the extended view "sale.res_config_settings_view_form" and that is why it is done in the module sale_*, in the section "Pricing". 
https://github.com/OCA/sale-promotion/blob/1331bce2984e48c8ce0555347ad3d03ee98897ec/sale_loyalty_partner_applicability/views/res_config_settings.xml#L5-L28
I had added a title as a help as they do in odoo but if it is more recommendable I add the help in the field itself.

chienandalu: Ok :) Add comment so there's no doubt in the future :D (the help string can be that one)

pilarvargas-tecnativa: done! :)

chienandalu (approved):

pedrobaeza:
Please use descriptive commit message. `add configuration option` is not descriptive. Which configuration option? Use instead (putting all the possible detail):
```
[IMP] loyalty_partner_applicability: Allow to share coupons inside same entity

There's a new configuration option in the settings to allow sharing coupons between members of the same trading entity.

It has to be in the base module for being shared between all the rest of applications, but each *_applicability module will contain extra bits of code to use it.
```

pilarvargas-tecnativa:
> Please use descriptive commit message. `add configuration option` is not descriptive. Which configuration option? Use instead (putting all the possible detail):

Thank you, I will improve my commits

Issue #169, opened 2023-10-31
Looks like coupon_domain_product_discount can not be used without sale_coupon_domain_product_discount


nivels-x1g1:
Looks like coupon_domain_product_discount can not be used without 

sale_coupon_domain_product_discount
```
"depends": ["sale_coupon", "coupon_domain_product_discount"],
```

but
coupon_domain_product_discount depends also from sale_coupon_domain_product_discount

```
coupon_domain_product_discount/models/coupon_program.py", line 25, in _filter_not_ordered_reward_programs
    domain = order._get_reward_values_discount_strict_limit_lines(program)
AttributeError: 'sale.order' object has no attribute '_get_reward_values_discount_strict_limit_lines'
````
## Module
coupon_domain_product_discount 
Version: 15.0.1.1.0

## Describe the bug
If Strict Per Product Limit was enabeld I get error above

## To Reproduce

1. create coupon program
2. set discount apply on to specific products
3. set discount apply on domain 
4. set strict per product limit

**Expected behavior**
no error
one module or definition of function also in coupon_domain_product_discount

chienandalu:
please check @pilarvargas-tecnativa

pilarvargas-tecnativa:
The issue is solved with the PR: https://github.com/OCA/sale-promotion/pull/172

PR #172, opened 2023-10-31
[15.0][FIX] sale_coupon_domain_product_discount, coupon_domain_product_discount: Fix migration

pilarvargas-tecnativa:
Fix migration of module sale_coupon_domain_product_discount. This module was split in v15 into coupon_domain_product_discount and s ale_coupon_domain_product_discount but the
_filter_not_ordered_reward_programs method of the coupon.program model should go in the sale_coupon_domain_product_discount module. In addition, the dependency of coupon_domain_product_discount has been corrected, since it is the base module and should depend on coupon and not on sale_coupon, and unusable data has been removed from the manifest.

Issue: https://github.com/OCA/sale-promotion/issues/169#issuecomment-1786770455

cc @Tecnativa TT45749


sale_coupon_domain_product_discount/__manifest__.py
ernesto-garcia-tecnativa, line 1: why remove the key "data"?, normally stay empty if not load xml or security files

pilarvargas-tecnativa: An empty key contributes nothing

ernesto-garcia-tecnativa: agree but i not understand why removed on this PR this is the reason for my question.

ernesto-garcia-tecnativa (approved):
Code and functional review ok

sale_coupon_domain_product_discount/__manifest__.py
pilarvargas-tecnativa: Because it is a general correction of the migration of the module that was split in two, there are changes in both modules correcting that migration and that is why I make that change as part of the correction.

chienandalu (approved):
Thanks!

PR #161, opened 2023-10-05
[16.0][MIG] coupon_mass_mailing: Migration to version 16.0

ernesto-garcia-tecnativa:
@tecnativa TT44320

PR #176, opened 2023-11-09
[16.0][MIG] coupon_mass_mailing > loyalty_mass_mailing: Renamed to loyalty_mass_mailing + Migration to version 16.0

pilarvargas-tecnativa:
Supersedes : https://github.com/OCA/sale-promotion/pull/161

There is nothing useful in the PR of the previous migration for this reason a new PR is made with a new migration from the beginning

cc @Tecnativa TT44320

@chienandalu @victoralmau please review

ping @pedrobaeza


loyalty_mass_mailing/models/loyalty_program.py
```diff
@@ -0,0 +1,68 @@
...
+            [("program_id", "in", self.ids)], ["program_id"], ["program_id"]
+        )
+        mapped_data = {m["program_id"][0]: m["program_id_count"] for m in mailing_data}
+        for program in self:
+            program.mailing_count = mapped_data.get(program.id, 0)
+
+    def action_mailing_count(self):
+        self.ensure_one()
+        xmlid = "mass_mailing.mailing_mailing_action_mail"
+        model = self.env["ir.model"].search([("model", "=", "res.partner")])
```
victoralmau:
```suggestion
        model = self.env["ir.model"].sudo().search([("model", "=", "res.partner")])
```

loyalty_mass_mailing/models/loyalty_program.py
```diff
@@ -0,0 +1,68 @@
...
+        partner_domains = [
+            ast.literal_eval(domain)
+            for domain in self.rule_ids.mapped("rule_partners_domain")
+            if domain
+        ]
```
pedrobaeza: You mustn't literally eval this. It breaks as sum as you have a variable in the right part. This is still not perfect if the domain is not normalized:

```suggestion
        domain_list = self.rule_ids.mapped("rule_partners_domain")
        partner_domains = "[" + "'|'," * (len(domain_list) - 1) + ",".join(x[1:-1] for x in domain_list) + "]"
```

Although you can create a computed field at coupon level getting this `partner_applicability_domain` for better encapsulating.

Replace `|` by `&` if all the conditions should be met.

pedrobaeza (changes requested):

victoralmau (approved):

loyalty_mass_mailing/models/loyalty_program.py
pilarvargas-tecnativa: I have opted for the computed field, as it seems clearer to me to obtain the necessary data. Thank you!

loyalty_mass_mailing/models/loyalty_program.py
pilarvargas-tecnativa: thanks :)

loyalty_mass_mailing/__manifest__.py
```diff
@@ -2,13 +2,13 @@
...
-    "version": "15.0.1.0.0",
+    "name": "Loyalty Mass Mailing",
+    "version": "16.0.1.0.0",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/sale-promotion",
     "license": "AGPL-3",
     "category": "Marketing",
-    "depends": ["sale_coupon", "mass_mailing"],
-    "data": ["views/mailing_mailing_view.xml", "views/sale_coupon_program_view.xml"],
+    "depends": ["loyalty", "loyalty_partner_applicability", "mass_mailing"],
```
chienandalu:
```suggestion
    "depends": ["loyalty_partner_applicability", "mass_mailing"],
```

loyalty_mass_mailing/models/loyalty_program.py
```diff
@@ -17,6 +18,24 @@ class SaleCouponProgram(models.Model):
...
+            domain_list = self.rule_ids.mapped("rule_partners_domain")
+            if any(domain.strip("[]") == "" for domain in domain_list):
+                program.partner_applicability_domain = "[]"
+            else:
+                program.partner_applicability_domain = (
+                    "["
+                    + "'|'," * (len(domain_list) - 1)
+                    + ",".join(x[1:-1] for x in domain_list)
+                    + "]"
+                )
```
chienandalu: Simpler:
```suggestion
            domain_list = [literal_eval(d) for d in self.rule_ids.mapped("rule_partners_domain") if d]
            program.partner_applicability_domain = str(expression.OR([d for d in domain_list]))
```

chienandalu (changes requested):

loyalty_mass_mailing/models/loyalty_program.py
chienandalu: > It breaks as sum as you have a variable in the right part

This will be user generated domains. Odoo itself discards the possibility of variable domains in core: https://github.com/odoo/odoo/blob/4ecb41c5f0fd15017ed837e9965ee8c84c77edee/addons/loyalty/models/loyalty_rule.py#L127

pedrobaeza: OK, if that's the case, then let's go with `literal_eval`, but after that, they have to be added with `expression.OR`

chienandalu: Yes, I proposed that before seeing your comment :sweat_smile:

loyalty_mass_mailing/models/loyalty_program.py
pedrobaeza: Do you need another comprehension list in the second line?
```
program.partner_applicability_domain = str(expression.OR([domain_list]))
```

chienandalu: Oh, yeah, it's useless :smile:

pilarvargas-tecnativa: done!

pilarvargas-tecnativa: > Do you need another comprehension list in the second line?

I don't think so...

loyalty_mass_mailing/models/loyalty_program.py
```diff
@@ -0,0 +1,79 @@
...
+    @api.depends("rule_ids.rule_partners_domain")
+    def _compute_partner_applicability_domain(self):
+        for program in self:
+            partner_domains = [
+                literal_eval(domain)
+                for domain in program.rule_ids.mapped("rule_partners_domain")
+                if domain
+            ]
+            if all(partner_domains):
+                program.partner_applicability_domain = expression.OR(partner_domains)
```
pedrobaeza, line 37: Shouldn't this be?

```suggestion
                program.partner_applicability_domain = expression.OR([partner_domains])
```

pilarvargas-tecnativa: Not necessary, here is an example of the value output in a test debug:
```
>>> partner_domains
[[('id', '=', 41)], [('id', '=', 42)]]

>>> program.partner_applicability_domain
"['|', ('id', '=', 41), ('id', '=', 42)]"
```

pedrobaeza (approved):

chienandalu (approved):
Tested and working! :+1:

PR #177, opened 2023-11-13
[16.0] [ADD] sale_loyalty_order_info: New module

marielejeune:
This module provides more info on the applied loyalty rewards on a sale order.
It provides:
* The total reward amount, computed tax included;
* A list of the applied promo codes;
* All generated coupons from this sale order;
* All programs applied to the sale order.


sale_loyalty_order_info/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+                order.promo_codes = str(codes)  # stock the list of codes in a string
+            else:
+                order.promo_codes = "[]"
+
+    @api.depends("order_line", "applied_coupon_ids", "code_enabled_rule_ids")
+    def _compute_programs(self):
+        self.program_ids = self.env["loyalty.program"]
+        for order in self:
+            # Consider claimable rewards
+            order._update_programs_and_rewards()
```
chienandalu: **question**: Won't this be triggering automatic updates or even double checking the rules.

sale_loyalty_order_info/models/sale_order.py
```diff
@@ -0,0 +1,65 @@
...
+            res = order._get_claimable_rewards()
+            for coupon, _ in res.items():
+                order.program_ids |= coupon.program_id
+
+            # Consider applied rewards
+            programs = order._get_reward_lines().mapped("reward_id.program_id")
```
chienandalu: **question**: maybe I don't understand what this method tries to achieve, but wouldn't it be enough to get:

```
order.order_line.reward_id.program_id + order.applied_coupon_ids.program_id + order.generated_coupon_ids.program_id
```
```

chienandalu (commented):
Thanks @marielejeune !

Some questions about the implemation:

sale_loyalty_order_info/models/sale_order.py
marielejeune: Hi @chienandalu 
You're completely right. In fact this code comes from custom code in version 14, and things were simplified since then.
`program_ids` must contain all applied programs (auto or triggered with code). In v16 Odoo creates a `sale.order.line` for each applied reward, so all applied programs are can simply be found in `order.order_line.reward_id.program_id`.
Changed the code.

sale_loyalty_order_info/models/sale_order.py
marielejeune: Changed in next version. Please see my other comment.

chienandalu (commented):
Where do the user get these infos. I can't see any related views

marielejeune:
> Where do the user get these infos. I can't see any related views

These fields are added for reporting purposes. One can add them in views if needed, but I didn't to avoid triggering computations (computed non stored fields).

chienandalu:
Ok, a note in the readme describing the current module purpose would be enough :slightly_smiling_face:

marielejeune:
> Ok, a note in the readme describing the current module purpose would be enough 🙂

Done!

chienandalu (approved):
Thanks :)

lmignon (approved):
LGTM (Code review)

PR #178, opened 2023-11-17
[16.0][FIX] sale_loyalty_partner_applicability: correct domain evaluation

pilarvargas-tecnativa:
This condition is erroneous since in case partner_domain is [ ] it will be evaluated as False and therefore will not fulfil the condition.

cc @Tecnativa TT44347


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #184, opened 2023-11-23
[16.0][MIG] coupon_criteria_multi_product>loyalty_criteria_multi_product: Migration to version 16.0

pilarvargas-tecnativa:
Supersedes: https://github.com/OCA/sale-promotion/pull/160

cc @Tecnativa TT44317

@chienandalu please review


loyalty_criteria_multi_product/migrations/pre-migration.py
```diff
@@ -0,0 +1,95 @@
...
+        """,
+    )
+    # coupon_criteria_ids to loyalty_criteria_ids
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE loyalty_criteria
+        ADD COLUMN IF NOT EXISTS rule_id INT
+        """,
+    )
```
chienandalu, line 35: Shouldn't this be done before?

loyalty_criteria_multi_product/migrations/pre-migration.py
```diff
@@ -0,0 +1,95 @@
...
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE loyalty_rule AS lr
+        SET loyalty_criteria = 'multi_product'
+        FROM loyalty_program AS lp
+        WHERE lr.program_id = lp.id
+        AND lp.coupon_criteria = 'multi_product'
+        """,
+    )
```
chienandalu, line 27: Shouldn't this be done after you insert the rules?

pilarvargas-tecnativa: You are right, thanks

loyalty_criteria_multi_product/migrations/pre-migration.py
```diff
@@ -0,0 +1,95 @@
...
```
chienandalu: How can this be if `loyalty_rule` is a new relation?

pilarvargas-tecnativa: The data corresponding to the rules are migrated to this table. 
https://github.com/OCA/OpenUpgrade/blob/5912a50fa93234fe37643c1091919f780d718dd1/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py#L39

chienandalu: Then why is it necessary to create them in the previous query?

pilarvargas-tecnativa: I have checked my notes again. In v15 "coupon_criteria" was defined at program level (coupon.program), now it has to be moved to rule level (loyalty.rule).  
v15: 
https://github.com/OCA/sale-promotion/blob/15.0/coupon_criteria_multi_product/models/coupon_program.py#L9

v16: https://github.com/OCA/sale-promotion/blob/60b86377e3b6c4de62086a2f266c0769c8111bd7/loyalty_criteria_multi_product/models/loyalty_rule.py#L9

That program could have or not defined rules that were stored in the "coupon_rule" table. 
v15: https://github.com/odoo/odoo/blob/15.0/addons/coupon/models/coupon_rules.py#L12-L19

If a program already has defined a rule, then there is already a record for that program and you have to adapt the data of the rule to consider "loyalty_criteria" but in case that program did not have other rules defined then you have to insert a new record to define the rule. So when re-analysing, the order of the queries is not relevant.

However, I have corrected the nomenclature of the field in loyalty_rule. Also the field loyalty_program.coupon_criteria does not have to be named when a new field is created in loyalty_rule.loyalty_criteria.

chienandalu: What I mean is that for every program there will be a rule as it's inherent to the way `coupon.rule` is binded to the `coupon.program` model. So when a program is created the rule is created and therefor it makes no sense to create a new one.

pilarvargas-tecnativa: Right, by default a record with the value "rule_min_quantity" is set even if there are no other rules set. I was thinking about the possibility of a program configured only with the multi-product criteria and that this would not affect the rules, but I have already verified that it does. Just update the record, no need to insert new records. Thank you.

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge nobump

loyalty_criteria_multi_product/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,32 @@
...
+#. Add as many criterias as needed for the desired promotion behavior.
+
+In the list of criterias we can configure:
+
+- Qty: The minimum quantity of products in the sale order to fulfill the condition.
+- Products: The products that should be present to fulfill the condition.
+
+Some examples:
+
+ ===== ================ ========
```
pedrobaeza: Don't put the initial space in all these lines

loyalty_criteria_multi_product/readme/CONTEXT.rst
```diff
@@ -0,0 +1,11 @@
+The functionality of the "Repeat" option that was included in this module is now included
+in the rules of an odoo promotion. In this functionality a criterion was established in
+which any of the products defined in the list of the criterion itself had to be in the
+sales order lines in the minimum quantity established for the promotion to be applied.
+This functionality can now be applied in the following way:
+- Defining or editing a promotion.
```
pedrobaeza, line 7: Put an empty line above this line.

loyalty_criteria_multi_product/readme/CONTEXT.rst
```diff
@@ -0,0 +1,11 @@
...
+- Modifying the rules to indicate a minimum quantity and selecting the products,
+  either by domain, specific products, product category or tag.
+This way the promotion will be applicable when any of those products are in the order l
```
pedrobaeza, line 11: Put an empty line above this line.

pedrobaeza (changes requested):

loyalty_criteria_multi_product/readme/CONTEXT.rst
pilarvargas-tecnativa: done, thanks!

pedrobaeza (approved):
Let's see if it's enough:

/ocabot merge nobump

PR #186, opened 2023-12-01
[16.0][FIX] loyalty_partner_applicability: Set default value for partner domain rule

pilarvargas-tecnativa:
It is necessary to set the default value, in this case "[ ]" to the rule_partners_domain field, overwriting the computed method "_program_type_default_values" that sets the default values depending on the type of program, because when creating a new program, these default rules are defined and the rule_partners_domain field, not being contemplated, its value will be "False", which cannot be interpreted as a correct domain. To avoid this error, the value is added to the rules defined by default.

The error:
image

cc @Tecnativa TT44344

@chienandalu @CarlosRoca13 please review

### DIFF
--- a/loyalty_partner_applicability/models/loyalty_rule.py
+++ b/loyalty_partner_applicability/models/loyalty_rule.py
@@ -1,6 +1,6 @@
 # Copyright 2023 Tecnativa - Pilar Vargas
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class LoyaltyRule(models.Model):
@@ -12,3 +12,11 @@ class LoyaltyRule(models.Model):
         help="Loyalty program will work for selected customers only",
         default="[]",
     )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        for vals in vals_list:
+            if not vals.get("rule_partners_domain", False):
+                vals["rule_partners_domain"] = "[]"
+        return res
--- a/sale_loyalty_partner_applicability/models/sale_order.py
+++ b/sale_loyalty_partner_applicability/models/sale_order.py
@@ -11,7 +11,7 @@ class SaleOrder(models.Model):
 
     def _get_partner_domain(self, rule, partner_id):
         domain = []
-        if rule.rule_partners_domain != "[]":
+        if rule.rule_partners_domain and rule.rule_partners_domain != "[]":
             allow_sharing = (
                 self.env["ir.config_parameter"].sudo().get_param("allow_coupon_sharing")
             )


loyalty_partner_applicability/models/loyalty_program.py
```diff
@@ -0,0 +1,21 @@
...
+        res = super()._program_type_default_values()
+        program_types_to_update = [
+            "promotion",
+            "loyalty",
+            "promo_code",
+            "buy_x_get_y",
+            "next_order_coupons",
+        ]
+        for program_type in program_types_to_update:
+            res[program_type]["rule_ids"][1][2].update({"rule_partners_domain": "[]"})
```
chienandalu: What if another module added a different structure?

pilarvargas-tecnativa: Possibility considered, thank you

chienandalu (commented):
Thinking it twice, would it be more resilient to override the create of the `program.rule` and simply add the default domain in the vals if not present?

pilarvargas-tecnativa:
> Thinking it twice, would it be more resilient to override the create of the `program.rule` and simply add the default domain in the vals if not present?

Yes, it's more resilient. Done :)

loyalty_partner_applicability/models/loyalty_rule.py
```diff
@@ -12,3 +12,11 @@ class LoyaltyRule(models.Model):
         help="Loyalty program will work for selected customers only",
         default="[]",
     )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        for vals in vals_list:
+            if "rule_partners_domain" not in vals or not vals["rule_partners_domain"]:
```
chienandalu:
```suggestion
            if not vals.get("rule_partners_domain"):
```

pilarvargas-tecnativa: thanks

chienandalu (approved):
:+1:

pedrobaeza:
I don't get why the default is not applied, even if coming from that "direct creation". Anyway, the way to fix this is to override `_program_type_default_values`, not to put this general "fallback". One can think that having this general fallback is better, but it's not because:

1. It's called everything you create a record, draining a bit the performance. It's not too much, but for a very corner case like this, it's not needed.
2. It can make debugging worse if we get a non-desired result, as the value assignation is not encapsulated and spread across several parts.

Anyway, I insist, why the default value is not applied? Is this bypassing the ORM somehow?

chienandalu:
> Anyway, I insist, why the default value is not applied? Is this bypassing the ORM somehow?

It's bypassed: https://github.com/odoo/odoo/blob/16.0/addons/loyalty/models/loyalty_reward.py#L20

pilarvargas-tecnativa:
ping @pedrobaeza

pedrobaeza (approved):
I have just added another safeguard on the sale part.

/ocabot merge patch

PR #187, opened 2023-12-05
[15.0][FIX] sale_coupon_criteria_order_based: Add tagged "post_install" to tests

pilarvargas-tecnativa:
This test requires a CoA to be installed, it should be tagged "post_install"

cc @Tecnativa TT46134

@pedrobaeza @CarlosRoca13 please review


pedrobaeza (approved):
My bad, the problem is this one, so let's fix the CI:

/ocabot merge patch

PR #181, opened 2023-11-22
[16.0][MIG] coupon_multi_gift>loyalty_multi_gift: Migration to version 16.0

pilarvargas-tecnativa:
Supersedes: https://github.com/OCA/sale-promotion/pull/162

cc @Tecnativa TT44321

@chienandalu please review

PR #196, opened 2023-12-11
[16.0][MIG] sale_coupon_multi_gift>sale_loyalty_multi_gift: Migration to version 16.0

pilarvargas-tecnativa:
Pendin from:
- [x] #181

cc @Tecnativa TT44348

@chienandalu please review


chienandalu (commented):
When there's only one multi-gift promo the wizard doesn't pop up and ll the products are put into the order, even all the options.

When there are several, the wizards shows up, but I can't do anything with the options. What's the plan?
image

pilarvargas-tecnativa:
> When there's only one multi-gift promo the wizard doesn't pop up and ll the products are put into the order, even all the options.

In v16, the promotions that have as a reward a product to choose between several products have this behaviour, if there are not several promotions the wizard does not open and they apply as a reward the default product.

Also, in v13 and v15, this is the behaviour, the wizard is opened by the module sale_coupon_order_suggestion (which is not yet migrated) with the gift icon in the order line that is part of a promotion. 

A test in v15, with the modules installed:
- sale_coupon_criteria_multi_product
- sale_coupon_multi_gift
- sale_coupon_selection_wizard
The wizard does not open and the default product of the promotion is applied.

A promotion is configured in which buying a white table or a black table you get a lamp or a bin as a reward.
image

Without sale_coupon_order_suggestion there is no icon to open the wizard and it does not open by itself.

v15

> When there are several, the wizards shows up, but I can't do anything with the options. What's the plan?

When there are alternative products as rewards, the list shows the configured products on the left side and on the right side you can select the product to apply.

Tests in v16:
A promotion is configured in which when the rules are met it will give away a bin or a lamp.
When there is only one promotion applicable, the default product (in this case the bin) will be given as a reward:

v16-1


When there are several promotions the wizard will open to select which promotion to apply and which product as a reward (same as in previous versions and same behaviour of odoo in v16):

v16-2

If there are several products as a reward, they will be shown in the wizard to choose whether to apply the default one or select another one. Each line of the list is a reward product, as I said before, on the left side the options are shown and on the right side the selected one.

Description of the rewards of the promotion (2 gifts):
image

List of products configured in the promotion (same as in previous versions - 2 gifts, 1 unit of one and 2 units of another):
image

List shown in the wizard (2 products with their options):
image

Once the promotion is applied in the sales order we have two rewards, 1 unit of bin and 2 units of table.
image

chienandalu:
- You're adding every option to the cart: that's not correct. You should add only the default one if you do it automatically
- The idea is to use the core wizard to hook our option choosing behavior so this module is now self containing.

pilarvargas-tecnativa:
@chienandalu I have relaunched

When the dependency is merged I will remove the commits that do not correspond to the migration.

chienandalu:
I'm getting errors when trying to apply it:

image

This is how it's configured
image

pilarvargas-tecnativa:
> I'm getting errors when trying to apply it:

Corrected! thanks

sale_loyalty_multi_gift/tests/test_sale_loyalty_multi_gift.py
```diff
@@ -104,3 +104,60 @@ def test_04_test_sale_coupon_test_multi_gift(self):
         self.assertEqual(2, discount_line_product_2.product_uom_qty)
         self.assertEqual(3, discount_line_product_4.product_uom_qty)
         self.assertFalse(discount_line_product_3)
+
+    def test_delete_01_loyalty_multi_gift_ids(self):
```
chienandalu: Put this test in your migration commit

pilarvargas-tecnativa: This commit is to remove it when the dependency is merged because it is not a test of this module.

chienandalu: Which one is it then? :thinking:

pilarvargas-tecnativa: This is the test you used to debug the duplicate records error in loyalty_multi_gift.

chienandalu: And why shouldn't we keep it as a regression test?

pilarvargas-tecnativa: But I think it would be more suitable to keep it in loyalty_multi_gift, don't you?

chienandalu: Ok, while it test that case :)

pilarvargas-tecnativa: done

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #197, opened 2024-01-16
[16.0][ADD] sale_loyalty_industry

santostelmo:
This module allow to restrict the industries in which the sale loyalty program applies.

Note: Moved from sale-workflow https://github.com/OCA/sale-workflow/pull/2859


cyrilmanuel (approved):
LGTM

cyrilmanuel:
Hi @pedrobaeza  can you make a review here please ?

pedrobaeza:
I think this module is not needed if you use `loyalty_partner_applicability` (and its sales counterpart `sale_loyalty_partner_applicability`, which is more generic to filter partners by any condition (including the industry), and it's also done at rule level, not program level, which is more inline with the new v16 approach.

imagen

I know yours is more "direct", but it's very limiting and in the end, I think using the generic one will give more versatility to your customer.

santostelmo:
> I think this module is not needed if you use `loyalty_partner_applicability` (and its sales counterpart `sale_loyalty_partner_applicability`, which is more generic to filter partners by any condition (including the industry), and it's also done at rule level, not program level, which is more inline with the new v16 approach.
> 
> imagen
> 
> I know yours is more "direct", but it's very limiting and in the end, I think using the generic one will give more versatility to your customer.

I have tested this module `sale_loyalty_partner_applicability` and indeed it seems to cover the needs. But i'm afraid that this domain is hard for a odoo user to build. It's less user friendly.

PR #205, opened 2024-03-14
[16.0][IMP] sale_loyalty_order_suggestion: Add hook method

pilarvargas-tecnativa:
Add hook method to help communicate this module with other modules that extend it by returning records from programs whose rules are based on products before filtering by order product and avoid extra modules that will not provide any functionality.

cc @Tecnativa TT48349

### DIFF
--- a/sale_loyalty_order_suggestion/models/sale_order.py
+++ b/sale_loyalty_order_suggestion/models/sale_order.py
@@ -73,19 +73,28 @@ class SaleOrder(models.Model):
             ]
         return programs.filtered(lambda p: p.id not in programs_to_delete)
 
-    def _available_programs(self):
-        self.ensure_one()
+    def _filter_programs_by_rules_with_products(self):
+        """Hook method. The objective of this method is to filter by rules that establish
+        products as criteria in a loyalty program and then in other methods make the
+        filtering that corresponds to each functionality."""
         valid_programs = self._get_available_programs()
         # Filters programs that have rules with minimum_qty > 0
         programs_with_minimum_qty = valid_programs.filtered(
             lambda x: any(rule.minimum_qty > 0 for rule in x.rule_ids)
         )
+        return programs_with_minimum_qty
+
+    def _available_programs(self):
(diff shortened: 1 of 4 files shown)


sale_loyalty_order_suggestion/models/sale_order.py
```diff
@@ -73,17 +73,24 @@ def _get_available_programs(self):
...
 
-    def _available_programs(self):
-        self.ensure_one()
+    def _filter_programs_by_rules_with_products(self):
+        """Hook method. The objective of this method is to filter by rules that establish
+        products as criteria in a loyalty program and then in other methods make the
+        filtering that corresponds to each functionality."""
         valid_programs = self._get_available_programs()
         # Filters programs that have rules with minimum_qty > 0
         programs_with_minimum_qty = valid_programs.filtered(
```
pedrobaeza, line 82: Why this module includes this minimum quantity criteria?

pilarvargas-tecnativa: The purpose of this module is to suggest promotions on order lines whose product is part of the rules of a promotion.

When the suggestion is made, it shows a gift icon which, when clicked, allows us to configure what is necessary for the promotion to be applicable (products and quantity to be added to the sales order).

By default, promotions in their rules consider the domain [ ] for the allowed products if a specific product is not set in their rules.

The rules of a promotion are not restrictive to each other but as soon as one is fulfilled, the promotion is applicable.

For this reason, and following the example of the DEMO promotion "Buy 3 large cabinets, get one for free" of the type "Buy X get Y", for a promotion to have its rules based on products, it is necessary to establish a quantity as the main criterion.

image

The rest of the types of promotions can also base their rules on products and that is why in this logic they are filtered by product quantity.

It should be clarified that the different types of promotions facilitate a more direct configuration (giving the configuration practically done) of the rules, but in reality we can configure anything as we need.

pedrobaeza (approved):

chienandalu (approved):

PR #206, opened 2024-03-15
[16.0][ADD] sale_loyalty_order_suggestion_multi_gift: New module

pilarvargas-tecnativa:
Pending from:

- [x] #205 
- [x] #196 
cc @Tecnativa TT48349

@chienandalu please review


pilarvargas-tecnativa:
ping @chienandalu

carlos-lopez-tecnativa:
I tested this, but I’m not sure if it’s working correctly.
My settings:

image
image
Despite these settings, I didn’t see the button on the sale order line
Could you confirm if this is the expected behavior?

image

pilarvargas-tecnativa:
> I tested this, but I’m not sure if it’s working correctly. My settings:

From the local tests I have done, it is possible that there is a conflict with one of the other installed modules. Some of them need bridging modules. I will look into it. If you test without setting a product in the rules it does suggest the promotion correctly. When you set the product it should also suggest the promotion and in fact in my local tests it does, I will investigate which module causes the conflict and if necessary resolve it.  Thanks

pilarvargas-tecnativa:
@carlos-lopez-tecnativa You can check again

carlos-lopez-tecnativa (approved):
Functionally tested, LGTM

chienandalu (approved):

PR #208, opened 2024-03-21
[16.0][MIG] sale_loyalty_auto_refresh: Migration to 16.0

nguyenminhchien:
Rename from `sale_coupon_auto_refresh` to `sale_loyalty_auto_refresh`


santostelmo (approved):

cyrilmanuel (approved):
LGTM

cyrilmanuel:
Hi @pedrobaeza can you merge this PR please ?

pedrobaeza:
/ocabot migration sale_loyalty_auto_refresh

@chienandalu do you see it correct?

sale_loyalty_auto_refresh/models/sale_coupon_refresh_mixin.py
```diff
@@ -21,15 +21,15 @@ def _get_auto_refresh_coupons_triggers(self) -> set:
...
+        the param key would be sale_loyalty_auto_refresh.sale_order_triggers,
         and the value something like: "warehouse_id,carrier_id".
 
         The method is overriden in the proper modules to set some basic triggers though
         """
         additional_triggers = (
             self.env["ir.config_parameter"]
             .sudo()
-            .get_param("sale_coupon_auto_refresh.%s_triggers" % (self._table), "")
+            .get_param("sale_loyalty_auto_refresh.%s_triggers" % (self._table), "")
```
chienandalu: chore: there should be a migration script renaming this parameter

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -42,8 +42,7 @@ def write(self, vals):
     def _auto_refresh_coupons(self):
         orders = self.filtered(type(self)._allow_recompute_coupon_lines)
         if orders:
-            orders = orders.with_context(skip_auto_refresh_coupons=True)
-            orders.recompute_coupon_lines()
+            orders.action_open_reward_wizard()
```
chienandalu: I'm not sure about this change of behavior. Why not using the new `_update_programs_and_rewards` method?

sale_loyalty_auto_refresh/tests/test_sale_loyalty_auto_refresh.py
```diff
@@ -0,0 +1,142 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo.tests import new_test_user, tagged
+
+from odoo.addons.sale_loyalty.tests.common import TestSaleCouponCommon
```
chienandalu: It'd be nice to have a diff here. Aside from that, at first glance I see lot of changes:

- Copyright and licences aren't respected from the original
- Tests cases are trimmed so you're not testing the same effects as before, so avoiding regression isn't granted or hard to tell...

chienandalu (changes requested):
Thanks for your work @nguyenminhchien :slightly_smiling_face: Just some comments:

sale_loyalty_auto_refresh/models/sale_order.py
nguyenminhchien: i tried calling it, but it didn't work as expected. anw, i made some changes to deal with the case there was more than 1 applicable promotion. please review.

sale_loyalty_auto_refresh/tests/test_sale_loyalty_auto_refresh.py
nguyenminhchien: Sorry about the Copyright part, i may remove it by mistake.
I just add 1 more unit test to check the case of having more than 1 applicable promotion.

cyrilmanuel:
Hello @chienandalu is it good for you  ? :)

sale_loyalty_auto_refresh/tests/test_sale_loyalty_auto_refresh.py
chienandalu, line 1: The test cases are completely changed and can't see a clear reason why...

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -40,10 +44,22 @@ def write(self, vals):
...
     def _auto_refresh_coupons(self):
-        orders = self.filtered(type(self)._allow_recompute_coupon_lines)
-        if orders:
-            orders = orders.with_context(skip_auto_refresh_coupons=True)
-            orders.recompute_coupon_lines()
+        self_ctx = self.with_context(skip_auto_refresh_coupons=True)
+        for order in self_ctx:
+            if order._allow_recompute_coupon_lines():
+                order._update_programs_and_rewards()
+                order.action_apply_rewards()
```
chienandalu: There's no real need to change the previous structure. Yo can:
```suggestion
        orders = self.with_context(skip_auto_refresh_coupons=True).filtered(type(self)._allow_recompute_coupon_lines)
        for order in orders:
            order._update_programs_and_rewards()
            order.action_apply_rewards()

File: sale_loyalty_auto_refresh/models/sale_order.py [hunk 5]
```diff
@@ -40,10 +44,22 @@ def write(self, vals):
...
+    def action_apply_rewards(self):
+        self.ensure_one()
+        claimable_rewards = self._get_claimable_rewards()
+        for coupon, reward in claimable_rewards.items():
+            try:
+                self._apply_program_reward(reward, coupon)
+                self._update_programs_and_rewards()
+            except Exception as e:
+                # Ignore exception errors to unblock the user when creating/writing
+                logger.warning(e)
```
chienandalu: This is going to be really noisy. I you need to do it, better just log it in debug mode.

chienandalu (changes requested):
Hi @nguyenminhchien some comments / doubts:

sale_loyalty_auto_refresh/tests/test_sale_loyalty_auto_refresh.py
nguyenminhchien: because the models of coupon and promotion were changed dramatically. so i trimmed the test to make it more focus on testing the applying of the promotion auto, than re-test the promotion flow, which has been done by sale_loyalty. anw, i'm going to roll it back.

cyrilmanuel (approved):

cyrilmanuel:
@chienandalu  is it good for you ?

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+    _inherit = ["sale.order.line", "sale.coupon.refresh.mixin"]
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        if self._check_skip_refresh():
+            return super().create(vals_list)
+
+        self_ctx = self.with_context(skip_auto_refresh_coupons=True)
+        lines = super(SaleOrderLine, self_ctx).create(vals_list)
+        lines.mapped("order_id")._auto_refresh_coupons()
```
simahawk:
```suggestion
        lines.order_id._auto_refresh_coupons()
```

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+            try:
+                self._apply_program_reward(reward, coupon)
+                self._update_programs_and_rewards()
+            except Exception as e:
+                # Ignore exception errors to unblock the user when creating/writing
+                logger.debug(e)
+
+    def _allow_recompute_coupon_lines(self):
+        """Returns whether reward lines in order ``self`` can be recomputed
+        automatically.
```
simahawk:
```suggestion
        """
        Check  if reward lines in ``self`` can be recomputed automatically.
```

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+        lines = super(SaleOrderLine, self_ctx).create(vals_list)
+        lines.mapped("order_id")._auto_refresh_coupons()
+        return lines
+
+    def write(self, vals):
+        if self._check_skip_refresh():
+            return super().write(vals)
+
+        old_data = self._read_recs_data()
+        old_orders = self.mapped("order_id")
```
simahawk:
```suggestion
        old_orders = self.order_id
```

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+        self_ctx = self.with_context(skip_auto_refresh_coupons=True)
+        res = super(SaleOrderLine, self_ctx).write(vals)
+        new_data = self._read_recs_data()
+        new_orders = self.mapped("order_id")
```
simahawk:
```suggestion
        new_orders = self.order_id
```

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+        new_triggers = self._new_trigger()
+        if old_data != new_data or any(x in new_triggers for x in vals):
+            (old_orders | new_orders)._auto_refresh_coupons()
+        return res
+
+    def unlink(self):
+        if self._check_skip_refresh():
+            return super().unlink()
+
+        orders = self.mapped("order_id")
```
simahawk:
```suggestion
        orders = self.order_id
```

sale_loyalty_auto_refresh/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,19 @@
...
+
+#. Go to *Settings > Technical > Config parameters*
+#. Add or update the key:
+
+   - For `sale.order`: `sale_loyalty_auto_refresh.sale_order_triggers`
+   - For `sale.order.line`: `sale_loyalty_auto_refresh.sale_order_line_triggers`
+#. In every add the fields seperated by commas that you want to add to the recomputation
+   triggers.
+
+⚠️ After configuring or removing a trigger a restart of Odoo is recommended so the
```
simahawk, line 17: I don't understand this statement. Either is required or is not. I'd remove it if not necessary because it's confusing.
I left a comment on the compute method.

sale_loyalty_auto_refresh/models/sale_coupon_refresh_mixin.py
```diff
@@ -0,0 +1,88 @@
...
+        additional_triggers = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("sale_loyalty_auto_refresh.%s_triggers" % (self._table), "")
+            .replace(" ", "")
+            .split(",")
+        )
+        return {x for x in additional_triggers if x in self._fields}
+
+    @api.depends(lambda self: list(self._get_auto_refresh_coupons_triggers()))
```
simahawk, line 38: Are we sure this complexity w/ the check on new triggers is necessary?
Data:
* the field you compute is non stored -> will get recomputed at every request and you can invalidate caching if needed
* any change on the config param won't have any effect on this depends
* in fact what you are doing is to load the triggers at need (eg: create/write)

Wouldn't be better to have an empty `api.depends`?

simahawk (approved):
Non blocking comments, only code review.
@chienandalu ping :)

santostelmo:
@chienandalu ping

santostelmo:
@SilvioC2C maybe you can help on the review / merge of this PR as it's been under review for some time.

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -40,10 +44,23 @@ def write(self, vals):
...
+            order.action_apply_rewards()
+
+    def action_apply_rewards(self):
+        self.ensure_one()
+        claimable_rewards = self._get_claimable_rewards()
+        for coupon, reward in claimable_rewards.items():
+            try:
+                self._apply_program_reward(reward, coupon)
+                self._update_programs_and_rewards()
+            except Exception as e:
```
rousseldenis: Isn't it a too broad except ?

nguyenminhchien: yes, but here, i want to ignore all exception errors to unblock the user when creating/writing

sale_loyalty_auto_refresh/migrations/post-migration.py
```diff
@@ -0,0 +1,18 @@
...
+
+from openupgradelib import openupgrade
+
+
+def update_parameter_key(env):
+    parameters = env["ir.config_parameter"].search(
+        [("key", "like", "sale_coupon_auto_refresh.")]
+    )
+    for parameter in parameters:
+        parameter.key = (parameter.key).replace(
```
SilvioC2C, line 11:
```suggestion
        parameter.key = parameter.key.replace(
```

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+            self._auto_refresh_coupons()
+        return res
+
+    def _auto_refresh_coupons(self):
+        orders = self.with_context(skip_auto_refresh_coupons=True).filtered(
+            type(self)._allow_recompute_coupon_lines
+        )
+        for order in orders:
+            order._update_programs_and_rewards()
+            order.action_apply_rewards()
```
SilvioC2C, line 53: Isn't `_update_programs_and_rewards()` already called by `action_apply_rewards()` internally? Do we need this double call?

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def action_apply_rewards(self):
+        self.ensure_one()
+        claimable_rewards = self._get_claimable_rewards()
+        for coupon, reward in claimable_rewards.items():
+            try:
+                self._apply_program_reward(reward, coupon)
+                self._update_programs_and_rewards()
+            except Exception as e:
```
SilvioC2C: Please catch any possible `psycopg2` exception and let them be raised, it'd be better to not ignore DB-level errors.

sale_loyalty_auto_refresh/models/sale_order.py
```diff
@@ -0,0 +1,153 @@
...
+            SaleOrder, self.with_context(skip_auto_refresh_coupons=True)
+        ).action_open_reward_wizard()
+
+    def _update_programs_and_rewards(self):
+        return super(
+            SaleOrder, self.with_context(skip_auto_refresh_coupons=True)
+        )._update_programs_and_rewards()
+
+
+class SaleOrderLine(models.Model):
```
SilvioC2C: Could be worth it to move `SaleOrderLine` to its own file `sale_order_line.py`.

SilvioC2C (approved):
Overall LGTM, a few comments here and there

sale_loyalty_auto_refresh/migrations/post-migration.py
nguyenminhchien: just curious but what's different or benefit of this suggested change?

sale_loyalty_auto_refresh/models/sale_order.py
nguyenminhchien: yes, we do. The first call to updates applied with the current state of the order. Then, after calling `_apply_program_reward()`, need to call it again.

sale_loyalty_auto_refresh/models/sale_order.py
santostelmo: Well I've seen different approaches regarding that. Some people prefer to keep it in the parent object file (`sale_order.py`). I am okay to keep like that. But move `SaleOrderLine` to his own file it's an option as well

sale_loyalty_auto_refresh/migrations/post-migration.py
SilvioC2C: Better readability :smile: parenthesis are redundant in this case

sale_loyalty_auto_refresh/models/sale_order.py
SilvioC2C: Yeah, usually if I have something like:
```
from odoo import fields, models


class MyModel(models.Model):
    _name = "my.model"

    name = fields.Char()
    line_ids = fields.One2many("my.model.line", "rec_id")


class MyModel(models.Model):
    _name = "my.model"

    descr = fields.Char()
    rec_id = fields.Many2one("my.model")
```
then it can make sense to have both classes in the same file.

But in this module's case, file's ~150 LoC already, and both classes contain business logic... might as well separate them for improved readability.

sale_loyalty_auto_refresh/models/sale_order.py
nguyenminhchien: anw, i updated to catch UserError, ValidationError only.

cyrilmanuel:
Hi @rousseldenis  is it good for you ?

santostelmo:
@simahawk is possible to merge that one ? 4 approvals

PR #209, opened 2024-03-27
[MIG] sale_loyalty_partner_applicability: Migration to 17.0

yannoliv:
Migrates loyalty_partner_applicability and sale_loyalty_partner_applicability. 

The only change from 16.0 to 17.0 is sale_loyalty_partner_applicability/views/res_config_settings.


chienandalu (commented):
Hi, @yannoliv Thanks for your contribution. As a general comment, you need to follow the migration guidelines so the module commit history is kept. You can check them here: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

yannoliv:
> Hi, @yannoliv Thanks for your contribution. As a general comment, you need to follow the migration guidelines so the module commit history is kept. You can check them here: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

Hi @chienandalu This is my first time migrating an OCA module. I've read through most of the conventions OCA has, but unfortunately, I do not seem to understand what I should do differently to keep the git history. Are you refering to this [code block](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#technical-method-to-migrate-a-module-from-160-to-170-branch)
)?

pedrobaeza:
Yes, that's it.

yannoliv:
Hi reviewers, I am really struggling with the `git am` command. Is it possible we should create a PR for each module? I think I understand the git history part as I now realize that with the `git am` command we preserve the whole history of changes of the module, instead of just adding it. What I still do not understand however is how I can do multiple at once?

pedrobaeza:
Following strictly the technical steps in the migration guide indicated above, which is the error you face?

yannoliv:
> Following strictly the technical steps in the migration guide indicated above, which is the error you face?

Hi, I do not experience any errors, but I am uncertain if my PR is done correctly.

pedrobaeza:
Well, here you haven't followed them. If you have another branch following them, just push it here.

blopa98-S73:
Hi @pedrobaeza , could you reopen this PR? Thanks

pedrobaeza:
It's not correct, without commit history, so no, it shouldn't be reopened.

PR #212, opened 2024-04-30
[ADD] loyalty_setup_validity

PlantBasedStudio:
Normally, the loyalty program is based on a fixed validity date. With this module, you
can choose between a fixed date or a date generated within a period of time after the
loyalty has been generated (for example, after a visit to the point of sale, a customer
can obtain a discount for 30 days).


pedrobaeza:
You must rebase the branch instead of doing a pull.

PR #217, opened 2024-06-06
[16.0][FIX] compute suggested_promotions when no product in so line

Kev-Roche:
In some cases, line of a sale order not have product_id and an error occurs for computing suggested_promotions field.


chienandalu (approved):
Thanks! 

cc @pilarvargas-tecnativa

pilarvargas-tecnativa (approved):
Thanks!

PR #221, opened 2024-07-13
[16.0 ][OU-FIX] loyalty_criteria_multi_product: column type + condition + criteria + repeat_product

pedrobaeza:
- It should be VARCHAR, not CHAR.
- Remove the condition for assigning the rule (as there can be m2m linked products although there criteria selection is another).
- Do the product m2m handle.
- Do the repeat_product handlement by SQL + fix it.

@Tecnativa


pedrobaeza:
Tested in real migration.

PR #228, opened 2024-10-21
[17.0][MIG] sale_coupon_criteria_order_based

mmequignon:


mmequignon:
ping @francesco-ooops @Ivorra78 @pilarvargas-tecnativa 
could you please drop a review ?
Thanks!

francesco-ooops (changes requested):
- Please check whether documentation is still accurate (eg: there's no "**Sales/Products/Promotion Programs** menu" in v17)

- Domain button is not visible: 

image

- Also, is it correct that "Based on order" domain is both visible in promotion form and "conditional rules" popup?

mmequignon:
> * Please check whether documentation is still accurate (eg: there's no "**Sales/Products/Promotion Programs** menu" in v17)

You're right, I totally forgot to update the doc


>     * Domain button is not visible:

Yes, adding it right now


> image
> 
>     * Also, is it correct that "Based on order" domain is both visible in promotion form and "conditional rules" popup?

Yes, and they are evaluated sequencially, which allows to have more complex rules.
1 domain which is common to all rules set on the program.
1 domain on the rule itself, which is specific to it.
With that we can imagine for a given program, the order should always be draft, 1 rule requires 10$ spent, the other a specific product to trigger the access to the reward.

francesco-ooops:
@mmequignon that's good, what about adding to the doc also this explanation about the presence of the field in both the promo and the condition rule to make more explicit how it's supposed to work?

mmequignon:
@francesco-ooops I updated the PR, can you please make another pass ?
Thanks !

francesco-ooops:
@mmequignon ping me after tests are fixed :)

mmequignon:
> @mmequignon ping me after tests are fixed :)

@francesco-ooops done

francesco-ooops:
@mmequignon mmm I don't see the field

image

mmequignon:
> @mmequignon mmm I don't see the field
> 
> image

That is because I added `base.group_no_one` (as it is done on product_domain) on the field.
you have to be have debug with assets enabled to see it

francesco-ooops:
Well, 

1) why?

2) that's the kind of stuff that makes functionals go to devs and ask "why don't I see this field?" and after their reply, "why isn't that in the documentation?"

I don't want to be rude, but our time is valuable too :)

mmequignon:
> Well,
> 
>     1. why?
> 
>     2. that's the kind of stuff that makes functionals go to devs and ask "why don't I see this field?" and after their reply, "why isn't that in the documentation?"
> 
> 
> I don't want to be rude, but our time is valuable too :)

I see your point, I just thought it would be better to stick to what was done for `product_domain`.
I'll drop this groups attribute.

sale_loyalty_criteria_order_based/models/loyalty_program.py
```diff
@@ -0,0 +1,34 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+from odoo.osv.expression import expression
+
+
+class LoyaltyProgram(models.Model):
+    _inherit = "loyalty.program"
+
+    rule_order_domain = fields.Char(string="Based on Order", default="[]")
```
pilarvargas-tecnativa, line 14: Where are you going to apply the criteria? At the rule level or at the program level?

sale_loyalty_criteria_order_based/models/loyalty_rule.py
```diff
@@ -0,0 +1,35 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+from odoo.osv.expression import expression
+
+
+class LoyaltyRule(models.Model):
+    _inherit = "loyalty.rule"
+
+    rule_order_domain = fields.Char(string="Based on Order", default="[]")
```
pilarvargas-tecnativa, line 15: Where are you going to apply the criteria? At the rule level or at the program level?

francesco-ooops (approved):
Functional ok!

francesco-ooops:
just a suggestion: UX-wise, it would be useful if domain condition would be displayed in the "condition" card, in order to make all the rules explicit (as it happens for other fields)

image

mmequignon:
> just a suggestion: UX-wise, it would be useful if domain condition would be displayed in the "condition" card, in order to make all the rules explicit (as it happens for other fields)
> 
> image

I'll add this to the ROADMAP

grindtildeath (approved):
LGTM :+1:

pilarvargas-tecnativa (changes requested):
The criteria cannot be applied at rule level and at program level. You have to decide at which level it should be applied and follow the odoo flow to apply a promotion and respect the changes that have been made since version 16.0 to adapt the promotion modules. You can check the 16.0 branch and see how other modules have been migrated for reference.

sale_loyalty_criteria_order_based/models/loyalty_program.py
grindtildeath, line 14: Wouldn't it make more sense to use a M2o on `ir.filter` in order to make it reusable between different programs?

grindtildeath:
@OCA/crm-sales-marketing-maintainers can you guys please reopen this one? :pray:

PR #248, opened 2025-01-23
[17.0][MIG] sale_loyalty_initial_date_validity: Migration to 17.0

BhaveshHeliconia:


sale_loyalty_initial_date_validity/__manifest__.py
```diff
@@ -3,11 +3,11 @@
...
     "name": "Sale Loyalty Initial Date Validity",
     "summary": "",
-    "version": "16.0.1.0.1",
+    "version": "17.0.1.0.0",
     "development_status": "Production/Stable",
     "category": "Sale",
     "website": "https://github.com/OCA/sale-promotion",
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "license": "AGPL-3",
-    "depends": ["sale_loyalty", "loyalty_initial_date_validity"],
```
victoralmau: Why do you remove this dependency?

victoralmau (changes requested):
Have you checked if this functionality already exists in core? 
I think not, `loyalty_initial_date_validity` is not needed anymore, and this module is not needed either. This migration and the 18.0 migration are not needed.

https://github.com/OCA/OpenUpgrade/pull/5168

PR #252, opened 2025-01-24
[18.0][MIG] loyalty_criteria_multi_product: Migration to 18.0

BhaveshHeliconia:


loyalty_criteria_multi_product/tests/test_loyalty_criteria_multi_product.py
```diff
@@ -1,20 +1,20 @@
...
+
+class TestLoyaltyCriteriaMultiProduct(BaseCommon):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
         product_obj = cls.env["product.product"]
         cls.pricelist = cls.env["product.pricelist"].create(
             {
-                "name": "Test pricelist",
+                "name": "Test Pricelist",
```
pilarvargas-tecnativa: This change only adds noise to the migration commit

loyalty_criteria_multi_product/tests/test_loyalty_criteria_multi_product.py
```diff
@@ -43,66 +43,70 @@ def setUpClass(cls):
...
-                                            (4, cls.product_b.id),
-                                            (4, cls.product_c.id),
-                                        ],
-                                    },
-                                ),
-                            ],
-                        },
-                    ),
-                    (
-                        0,
```
pilarvargas-tecnativa: Why is this second rule being eliminated?

loyalty_criteria_multi_product/tests/test_loyalty_criteria_multi_product.py
```diff
@@ -43,66 +43,70 @@ def setUpClass(cls):
...
+                                            Command.link(cls.product_c.id),
                                         ],
                                     },
                                 ),
                             ],
                         },
-                    ),
-                ],
-                "reward_ids": [
-                    (
```
pilarvargas-tecnativa: And the rewards?

pilarvargas-tecnativa (changes requested):

pilarvargas-tecnativa:
image
I can't add the products.

BhaveshHeliconia:
@pilarvargas-tecnativa : In v16, you can add products because the sale_loyalty_criteria_multi_product module is installed, which grants access rights to the loyalty.criteria model.
In v18 runboat, this module is not installed, so you can’t add products. This happens because the loyalty_criteria_multi_product module itself doesn’t provide access rights for loyalty.criteria—those rights come from sale_loyalty_criteria_multi_product.
I tested locally with both modules installed in v18, and I was able to add products successfully.

-  Screenshot from 2025-08-14 10-52-36

pilarvargas-tecnativa (approved):
Thanks!

@Tecnativa TT54442

@carlos-lopez-tecnativa please review

loyalty_criteria_multi_product/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_loyalty_criteria,access_loyalty_criteria,model_loyalty_criteria,base.group_user,0,0,0,0
```
carlos-lopez-tecnativa, line 2: Why are all access rights disabled with 0?

I cannot create records in this model.
image

carlos-lopez-tecnativa (changes requested):

loyalty_criteria_multi_product/security/ir.model.access.csv
carlos-lopez-tecnativa: > In v16, you can add products because the sale_loyalty_criteria_multi_product module is installed, which grants access rights to the loyalty.criteria model. In v18 runboat, this module is not installed, so you can’t add products. This happens because the loyalty_criteria_multi_product module itself doesn’t provide access rights for loyalty.criteria—those rights come from sale_loyalty_criteria_multi_product. I tested locally with both modules installed in v18, and I was able to add products successfully.

Reading this comment, I see the `ir.model.access.csv` but I don’t understand why this module cannot add the access rights without depending on another module. I think it would be better if this module added the security rights. What do you think? Any reason not to do this?

BhaveshHeliconia: @carlos-lopez-tecnativa : You can check the problem I explained above with pilarvargas-tecnativa.

carlos-lopez-tecnativa: > @carlos-lopez-tecnativa : You can check the problem I explained above with pilarvargas-tecnativa.

I read your comment where you said that the module `sale_loyalty_criteria_multi_product` is the one that adds the access right, but my question was: why doesn’t this module add the access rights itself?

Now I understand that this was done intentionally to avoid adding an explicit dependency on `sales_team`. Personally, I would prefer that this module either added the access right itself or at least indicated this somewhere in the README, but anyway, let’s move forward.

carlos-lopez-tecnativa (approved):

PR #251, opened 2025-01-24
[18.0][MIG] coupon_chatter: Migration to 18.0 

BhaveshHeliconia:

PR #254, opened 2025-01-24
[18.0][MIG] loyalty_limit: Migration to 18.0

BhaveshHeliconia:


loyalty_limit/i18n/es.po
```diff
@@ -1,173 +0,0 @@
-# Translation of Odoo Server.
```
victoralmau: You should not remove any translations in the migration (similar to https://github.com/OCA/sale-promotion/pull/251#discussion_r2169593798)

loyalty_limit/tests/test_loyalty_limit.py
```diff
@@ -0,0 +1,271 @@
...
+                        },
+                    )
+                ],
+            }
+        )
+
+    def test_salesmen_limit_fields_default_computation(self):
+        """Test that the default computation sets limit counts to 0."""
+        program = self.promotion_with_salesman_limit
+        program._compute_salesmen_limit_count()
```
victoralmau:
```suggestion
        program._compute_salesmen_limit_count()
```
It is not necessary to call the compute manually, the value should already be appropriate.

loyalty_limit/tests/test_loyalty_limit.py
```diff
@@ -0,0 +1,271 @@
...
+            "Expected salesmen_limit_count to be 0 by default",
+        )
+        self.assertEqual(
+            program.salesmen_limit_used_count,
+            0,
+            "Expected salesmen_limit_used_count to be 0 by default",
+        )
+
+        for limit in program.salesmen_limit_ids:
+            limit._compute_times_used()
```
victoralmau: It is not necessary to call the compute manually, the value should already be appropriate.

victoralmau (changes requested):

BhaveshHeliconia:
@victoralmau : Please review. Thanks

victoralmau:
Oops, I think you made a mistake adding an extra commit, you need to add those changes only to the migration commit of this module.

ejemplo

BhaveshHeliconia:
@victoralmau : Oops, you're right—thanks for catching that. I've removed the extra commit. Let me know if I missed anything else.

victoralmau (approved):
Code review OK

Ping @pilarvargas-tecnativa

pilarvargas-tecnativa (approved):

PR #261, opened 2025-01-24
[18.0][MIG] sale_loyalty_initial_date_validity: Migration to 18.0

BhaveshHeliconia:


victoralmau (changes requested):
Similar comment to https://github.com/OCA/sale-promotion/pull/248#pullrequestreview-3015409789

BhaveshHeliconia:
@victoralmau : I checked in the core, and you're right—this module is no longer needed. Thanks for pointing it out!

PR #265, opened 2025-01-27
[16.0] improves loyalty_partner_applicability

### LINKED ISSUE #238: [16.0] Improve partner applicability on loyalty programs
# Context
The OCA addons `loyalty_partner_applicability` and `sale_loyalty_partner_applicability` allow you to restrict the applicability of loyalty programs to specific partners.  This is useful when you want to apply a loyalty program only to a subset of your partners. In the current implementation, the restriction is based on a domain applicable to the partners specified at the loyalty rule level. When the system check if a program can be applied to a sale order, it search for all the valid programs and then iterate on each rule to check if the partner is applicable. The program is therefore applicable if at least one rule is applicable to the partner of the sale order.
Moreover, rules not applicable to the partner are also applied as long as at least one rule is applicable for the same program. Without knowing the motivation behind the current implementation, this looks like a strange behavior to me.

# Proposal
I would change the implementation to allow to specify the partner applicability at the program level. This would allow to have a more efficient implementation to check if a program is applicable to a sale order and also simplify the configuration by enforcing the concept of partner applicability at the program level. This change will not change the behavior of the current implementation
but place the concept of partner applicability at the right level according to the current behavior.

In the same time, I would also allow to restrict the applicability to a specific list of partners. Sometimes, you want to apply a program to a specific set of partners that can't be defined by a domain.

What do you think about this change? @Tecnativa @pilarvargas-tecnativa  @chienandalu @ernesto-garcia-tecnativa @pedrobaeza

lmignon:
fixes #238

* [x] includes #267


pedrobaeza:
Sorry, I haven't been able yet to check this. In principle, I agree it's weird to select the partner domain at a sublevel. Maybe a trick is to set it at "header", and propagate on the background to each rule, but as said, I'm only talking from the very surface.

lmignon:
Thank you for your comment @pedrobaeza . I share the same feeling as yours and can't imagine a valid use case to keep the restriction at the rule level. I keep it at the rule level to ensure the backward compatibility but it comes with a high performance cost. I would really like to remove it from the rule level and only keep it a program level. In this way the restriction could be part of the domain used to search for applicable programs. It will be more efficient.

pilarvargas-tecnativa:
@lmignon Can you do a rebase and push again to check in runbot, please?

loyalty_partner_applicability/models/loyalty_partner_applicability_mixin.py
```diff
@@ -0,0 +1,110 @@
...
+from odoo.tools.safe_eval import datetime, safe_eval
+
+
+class LoyaltyPartnerApplicabilityMixin(models.AbstractModel):
+
+    _name = "loyalty.partner.applicability.mixin"
+    _description = "Manage the applicability of loyalty elements to partners"
+
+    partner_ids = fields.Many2many(
+        string="Specific customers",
```
pilarvargas-tecnativa: It is best to avoid ‘Specific customers’ because it is ambiguous and does not clearly reflect that the field limits the promotion to certain customers, which may confuse the user about its purpose.

```suggestion
        string="Allowed partners",
```

loyalty_partner_applicability/models/loyalty_partner_applicability_mixin.py
```diff
@@ -0,0 +1,110 @@
...
+        comodel_name="res.partner",
+        help="This field allow to restrict the applicability to a specific set of partners",
```
pilarvargas-tecnativa:
```suggestion
        help="Only the selected partners will be eligible for this promotion.",
```

loyalty_partner_applicability/models/loyalty_partner_applicability_mixin.py
```diff
@@ -0,0 +1,110 @@
...
+        default=lambda p: p.env.context.get("default_partner_ids"),
+    )
+
+    partner_domain = fields.Char(
+        string="Specific customers domain",
```
pilarvargas-tecnativa:
```suggestion
        string="Allowed partners domain",
```

loyalty_partner_applicability/models/loyalty_partner_applicability_mixin.py
```diff
@@ -0,0 +1,110 @@
...
+        help="Applicable to the customers that match the domain",
```
pilarvargas-tecnativa:
```suggestion
        help="Define the domain to restrict which partners are eligible for this promotion.",
```

loyalty_partner_applicability/models/loyalty_rule.py
```diff
@@ -1,22 +1,15 @@
...
-    @api.model_create_multi
-    def create(self, vals_list):
-        res = super().create(vals_list)
-        for vals in vals_list:
-            if not vals.get("rule_partners_domain", False):
-                vals["rule_partners_domain"] = "[]"
-        return res
+class LoyaltyRule(models.Model):
+    _name = "loyalty.rule"
+    _inherit = ["loyalty.rule", "loyalty.partner.applicability.mixin"]
```
pilarvargas-tecnativa: If the purpose of refactoring is to apply the rule at the promotion level, I find it a little confusing that a mixin is created and that it is still possible to apply this at the rule level. Ultimately, what is the approach? If it is applied at the promotion level, it would have to be removed at the rule level, as it could define different clients at two different levels, but only one would have priority, that of the rule. This would cause errors when configuring the rule and unexpected results when applying the promotion.

pilarvargas-tecnativa (changes requested):

eugenios73:
Hello @lmignon when you solve the comments, tell me and I'll review the functionality.

loyalty_partner_applicability/models/loyalty_rule.py
lmignon: @pilarvargas-tecnativa I kept the definition at rule level to not break the existing implementation and don't introduce a change of behavior in used version.

lmignon:
> Hello @lmignon when you solve the comments, tell me and I'll review the functionality.

@eugenios73  comment solved or responded

loyalty_partner_applicability/models/loyalty_rule.py
pilarvargas-tecnativa: But it is understood that refactoring brings about this change in behaviour to configure the rules at the promotion level rather than at the rule level. If the current behaviour is to be maintained, then it is not refactoring. It is adding something new and changing the names of what already exists. Correct me if I am wrong.

pilarvargas-tecnativa: I believe that in this case, maintaining both systems seems unnecessary and can be completely simplified. One option would be to replace the old logic by doing an OR in the migration script to manage the different partner domains for each rule. This would maintain data compatibility without complicating the active code.
My recommendation would be to apply this simplification to avoid future problems and facilitate maintenance.

pilarvargas-tecnativa:
Hello @lmignon 

Any news regarding my comments? Would you prefer me to continue with the refactoring as I suggested? Or would you prefer to finish this refactoring yourself?

lmignon:
@pilarvargas-tecnativa The problem if we remove the applicability at rule level  could be that you will break existing implementations on v16.

pilarvargas-tecnativa:
> @pilarvargas-tecnativa The problem if we remove the applicability at rule level could be that you will break existing implementations on v16.

A script would need to be added to migrate what is established in the rules to the field added in the promotion, combining it with an OR would be sufficient.

The rest is to remove the logic for the rule that has no effect.

lmignon:
> > @pilarvargas-tecnativa The problem if we remove the applicability at rule level could be that you will break existing implementations on v16.
> 
> A script would need to be added to migrate what is established in the rules to the field added in the promotion, combining it with an OR would be sufficient.
> 
> The rest is to remove the logic for the rule that has no effect.

@pilarvargas-tecnativa If a program is configured with applicability at rule level, the removing of this functionality will lead to cases where customers could wrongly benefit from loyalties  after the update of this addon on 16.0. I'm really concerned by not introducing unwanted behavior on 16.0...

pilarvargas-tecnativa:
> @pilarvargas-tecnativa If a program is configured with applicability at rule level, the removing of this functionality will lead to cases where customers could wrongly benefit from loyalties after the update of this addon on 16.0. I'm really concerned by not introducing unwanted behavior on 16.0...

This is why I mentioned that a migration script must be included to transfer any domains configured in the rules to the new field.

The script must consist of going through the rules, checking for domains, and if there is more than one rule with that domain set, combining them with an OR in the new field.

With this script, the behaviour will be the same when deleting the field from the rules.

lmignon:
@pilarvargas-tecnativa If you have 2 rules for the same product with a different amount for different parnters and you combine the the domain a the program level, each partner will benefit of the 2 rules in place of only one.

pilarvargas-tecnativa:
> @pilarvargas-tecnativa If you have 2 rules for the same product with a different amount for different parnters and you combine the the domain a the program level, each partner will benefit of the 2 rules in place of only one.

But promotions don't work that way. When there are several rules, as soon as one is met, the promotion is applicable. In other words, if there are two rules for the same product and different partners, it will not discriminate that the rule is met in its entirety, but as soon as the least restrictive condition is met, the rule will be applicable to both partners. You can test this and verify that the domain will function as an OR.  Outside of the partner's applicability, you can configure a rule in which you buy one product (regardless of amount) and another rule in which you buy two units of another product.

When you apply the promotion at the moment when there is one quantity of either of the two products added to different rules, the promotion is applicable, and the same thing happens within the same rule by putting one quantity with one of the two products added to different rules. When you apply the promotion at the moment when there is 1 quantity of either of the two products added to different rules, the promotion is applicable, and the same thing happens within the same rule by putting 1 quantity with X amount; the first condition that is met will apply the rule.

lmignon:
I'll change the code, In any case, I will not be affected.

lmignon:
@pilarvargas-tecnativa I made the change, but it seems to me that it still causes a change in behavior (I'm really not convinced it's a good idea since we are loosing flexibility and we don't get a real benefit into the code). If today there is a loyalty program with a coupon for customer X and a rule for customer Y, with the change, after the update of the addon, customer Y can use the coupon and the original UC can no longer be configured.

pilarvargas-tecnativa:
> @pilarvargas-tecnativa I made the change, but it seems to me that it still causes a change in behavior (I'm really not convinced it's a good idea since we are loosing flexibility and we don't get a real benefit into the code). If today there is a loyalty program with a coupon for customer X and a rule for customer Y, with the change, after the update of the addon, customer Y can use the coupon and the original UC can no longer be configured.

However, this has not yet been merged, so no one should be using it; there will only be promotions configured at the rule level. When the change is rolled out, the promotion should continue to behave in the same way. In any case, it is not correct to have applicability configured at both levels. And in the hypothetical case that someone has rolled out this PR, if it is more restrictive at the promotion level, then the rule domain should be removed.

pilarvargas-tecnativa:
In my local tests:

- I have set up a local database with the above prior to refactoring.
- I have created a promotion with two rules for two different partners and different minimum amounts. The result of applying this promotion is that when the order meets the lowest amount of the rules, it is applicable to either of the two partners.
- I added the PR to my local instance and updated the module to run the migration scripts.
- The domain result is correct; the promotion should be applicable to both partners and therefore the domains have been combined correctly.

On the other hand, the promotion is not being restricted to other partners as expected. The promotion should only be applicable to the two partners in the rules and not to the rest.

pedrobaeza:
I'm not sure I'm getting the problem depicted by @lmignon, but he is talking about coupons, not programs. Anyway, I would say once the coupon is issued, it has the partner assigned, so no possibility of being mismatched, isn't it?

pilarvargas-tecnativa:
> I'm not sure I'm getting the problem depicted by @lmignon, but he is talking about coupons, not programs. Anyway, I would say once the coupon is issued, it has the partner assigned, so no possibility of being mismatched, isn't it?

Coupons follow a different path.

Gift cards and e-wallets have no rules. For ‘coupon’ promotions that generate coupons, you have to do it manually by setting up the partner, and then there is the promotion that generates a coupon for the next order, which follows the same application rules as all other promotions. If the conditions are met, the coupon for the next order is generated, so the focus is just on migrating the domain correctly, and that’s it.

All that remains is to adjust the application logic so that the promotion is restricted to that domain for the order.

pedrobaeza:
So is there a problem or not with this change?

sale_loyalty_partner_applicability/models/sale_order.py
```diff
@@ -1,77 +1,49 @@
...
-                res[program] = {
-                    "error": _("The customer doesn't have access to this reward.")
-                }
-        return res
+        self.ensure_one()
+        # We add the order to the context to be able to filter the rules based on the
+        # partner domain.
+        # This is needed to check the partner of the order and filter the rules based
+        # on the partner domain.
+        return super()._program_check_compute_points(programs.with_context(order=self))
```
pilarvargas-tecnativa, line 33: > So is there a problem or not with this change?

The logic would need to be adjusted so that it only applies to the domain, and the scripts would be correct.

pedrobaeza:
Then please suggest the needed changes.

pilarvargas-tecnativa (approved):
I debugged locally and the reason why the restriction was not being applied was due to a problem with the dirty database. I ran the local tests again with everything complete and everything is correct. It is also covered in the added tests.    

LGTM, thanks!

victoralmau (approved):
Code review OK. Although it may be difficult to review, it will now be more consistent.

lmignon:
Thank you to all for your reviews and suggestions. I still think we should have keep the ability to restrict the applicability to partners also at rule level. We could imagine UC where for a same program the reward points for the same product could depends on the partners.

pedrobaeza:
@lmignon but AFAIK, that reward part is not tied to the applicability rules, so keeping the partner applicability at that level doesn't improve anything. Can you expand that use case?

lmignon:
> @lmignon but AFAIK, that reward part is not tied to the applicability rules, so keeping the partner applicability at that level doesn't improve anything. Can you expand that use case?

@pedrobaeza The number of point rewarded and collected for a product is defined on the rule. 
Imagine the case where for a same program you want to reward customers who have subscribed to a subscription more generously than those without a subscription for the same product.
With the ability to also define the applicability at rule level you can define your program as

Loyalty Program A:
* Rule 1: Product1, reward point 10, domain for Partners with a subscription
* Rule 2: Product1, reward point 5, domain for Partners without subscription

This kind of configuration is no more possible with the last requested change.

pilarvargas-tecnativa:
> @pedrobaeza The number of point rewarded and collected for a product is defined on the rule. Imagine the case where for a same program you want to reward customers who have subscribed to a subscription more generously than those without a subscription for the same product. With the ability to also define the applicability at rule level you can define your program as

But that's not really how it works. The rules are combined, not restricted. In your use case:

Loyalty Program A:

- Rule 1: Product1, reward point 10, domain for partner1 
- Rule 2: Product1, reward point 5, domain for partner2 

The result would be the same for both partner1 and partner2 in an order with product1, reward points = rule1+rule2 = 15

lmignon:
> > @pedrobaeza The number of point rewarded and collected for a product is defined on the rule. Imagine the case where for a same program you want to reward customers who have subscribed to a subscription more generously than those without a subscription for the same product. With the ability to also define the applicability at rule level you can define your program as
> 
> But that's not really how it works. The rules are combined, not restricted. In your use case:
> 
> Loyalty Program A:
> 
>     * Rule 1: Product1, reward point 10, domain for partner1
> 
>     * Rule 2: Product1, reward point 5, domain for partner2
> 
> 
> The result would be the same for both partner1 and partner2 in an order with product1, reward points = rule1+rule2 = 15



> > @pedrobaeza The number of point rewarded and collected for a product is defined on the rule. Imagine the case where for a same program you want to reward customers who have subscribed to a subscription more generously than those without a subscription for the same product. With the ability to also define the applicability at rule level you can define your program as
> 
> But that's not really how it works. The rules are combined, not restricted. In your use case:
> 
> Loyalty Program A:
> 
>     * Rule 1: Product1, reward point 10, domain for partner1
> 
>     * Rule 2: Product1, reward point 5, domain for partner2
> 
> 
> The result would be the same for both partner1 and partner2 in an order with product1, reward points = rule1+rule2 = 15

With my code, the rule1 will not be applied on order from partner2 and vice versa.

pedrobaeza:
But that means that this module is also changing other part of the `loyalty` behavior, not the partner applicability, so that should go in another module, not here. I think this module should be kept simple, and let the lines part on an extra module on top of it. It was your initial argumentation for the refactorization: simplify the code and the user experience. In current form, I think it's the contrary, having more code and possible combinations that leds to user confusion.

lmignon:
@pedrobaeza I made all the changes requested even the one I'm not convinced it should be done. I improved the way you could define your restriction on partners, and the use of the mixin before the last change makes the code easy to understand and to maintain. I also added a lot of tests for all the UCs. The day I need to allows to restrict partner at rule level, the best way to do it is to add a mixin as it was done to share the logic between the program and the rule... I hope this PR should at least be merged as is.

pedrobaeza:
OK, don't worry, I don't argue more, so let's merge it.

/ocabot merge major

PR #282, opened 2025-05-05
[16.0][FIX-I18N] sale_loyalty_order_suggestion_multi_product: Correct translation of "and" in loyalty rule description

pilarvargas-tecnativa:
cc @Tecnativa TT56160

@pedrobaeza please review

### DIFF
--- a/sale_loyalty_order_suggestion_multi_product/wizard/sale_loyalty_reward_wizard.py
+++ b/sale_loyalty_order_suggestion_multi_product/wizard/sale_loyalty_reward_wizard.py
@@ -53,14 +53,17 @@ class SaleLoyaltyRewardWizard(models.TransientModel):
 
     def _compute_loyalty_rule_line_description(self):
         self.loyalty_rule_line_description = False
-        products = list(
-            set(
-                self.selected_reward_id.program_id.rule_ids.loyalty_criteria_ids.product_ids
-            )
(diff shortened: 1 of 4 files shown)


sale_loyalty_order_suggestion_multi_product/wizard/sale_loyalty_reward_wizard.py
```diff
@@ -60,7 +60,12 @@ def _compute_loyalty_rule_line_description(self):
         )
         if self.selected_reward_id and not self.applicable_program and products:
             if len(products) > 1:
-                products_str = f"{', '.join(map(str, [product.name for product in products][:-1]))} {_('and')} {[product.name for product in products][-1]}"  # noqa: B950
+                product_names = [product.name for product in products]
```
pedrobaeza:
```suggestion
                product_names = products.mapped("name")
```

And previously, `self.selected_reward_id.program_id.rule_ids.loyalty_criteria_ids.product_ids` already returns not duplicated products, so no need of `set` and then `list`.

pilarvargas-tecnativa: done, thanks

pedrobaeza (approved):
/ocabot merge patch

PR #284, opened 2025-06-20
[14.0][FIX] sale_coupon_default_product: Avoid duplicates in group-by based on discount_line_product_id field

Chionne27:
Before this fix the product linked to the discount line product id field in a coupon program was duplicated generating different groups with the same product name in list view

### DIFF
--- a/sale_coupon_free_shipping_default_product/__manifest__.py
+++ b/sale_coupon_free_shipping_default_product/__manifest__.py
@@ -1,6 +1,7 @@
 {
     "name": "Sale coupon free shipping default product",
-    "summary": "avoids creating multiple free shipping products",
+    "summary": """avoids creating multiple free shipping products and
+    creating multiple discount products if a coupon program is duplicated""",
     "version": "14.0.1.0.0",
     "category": "Hidden/Tools",
     "author": "Ooops, Odoo Community Association (OCA)",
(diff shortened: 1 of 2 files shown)


SirPyTech (changes requested):
Thanks for the PR!
Renaming a module within a stable version will break any other module depending on this one and, depending on the data, even the DB.

It can be done in some specific cases (for instance in case of module name conflict), but I think it's better to avoid it.
Is there a specific reason why it is needed?

It's better to rename modules within a major version migration, where **OpenUpgrade** can help (ref. https://oca.github.io/OpenUpgrade/use_cases/module_renaming.html).

Chionne27:
> Thanks for the PR! Renaming a module within a stable version will break any other module depending on this one and, depending on the data, even the DB.
> 
> It can be done in some specific cases (for instance in case of module name conflict), but I think it's better to avoid it. Is there a specific reason why it is needed?
> 
> It's better to rename modules within a major version migration, where **OpenUpgrade** can help (ref. https://oca.github.io/OpenUpgrade/use_cases/module_renaming.html).

@SirPyTech  the rename module could be necessary because i added a second functionality over the default shipping product default which prevent to duplicate a default product of a coupon program when the latter one is duplicated and also the duplication of coupons program's group in the list view.

I made a commit that could allow to process the module rename. Could you check if everything is right or if something is missing?

SirPyTech:
Moved to https://github.com/OCA/sale-promotion/pull/290.

PR #290, opened 2025-08-04
[ADD] sale_coupon_default_product: avoid creating multiple product if a coupon program is duplicated

Chionne27:
Before this new module the product linked to the discount line product id field in a coupon program was duplicated generating different groups with the same product name in list view


francesco-ooops (approved):
Functional ok!

sale_coupon_default_product/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This modules allows to avoid creating multiple product if a coupon program is duplicated.
```
SirPyTech: **typo**: 
```suggestion
This module allows to avoid creating multiple product if a coupon program is duplicated.
```

sale_coupon_default_product/models/coupon_program.py
```diff
@@ -0,0 +1,14 @@
+from odoo import api, models
+
+
+class CouponProgram(models.Model):
+    _inherit = "coupon.program"
+
+    @api.model
+    def create(self, vals):
+        # avoids creating multiple products if a coupon program is duplicated
+        if self.discount_line_product_id:
```
SirPyTech: **suggestion**: Please do not use the values of fields in `self` because `create` is a `model` method, so the content of `self` should not be relevant:
> Decorate a record-style method where ``self`` is a recordset, but its contents is not relevant, only the model is.

(from https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/odoo/api.py#L309-L310)

sale_coupon_default_product/i18n/it.po
```diff
@@ -0,0 +1,59 @@
...
+#: model:ir.model,name:sale_coupon_default_product.model_coupon_program
+msgid "Coupon Program"
+msgstr "Programma coupon"
+
+#. module: sale_coupon_default_product
+#: model:ir.model.fields,field_description:sale_coupon_default_product.field_res_company__free_shipping_product_default
+#: model:ir.model.fields,field_description:sale_coupon_default_product.field_res_config_settings__free_shipping_product_default
+msgid "Default promo product for free shipping"
+msgstr "Prodotto promozionale predefinito per spedizione gratuita"
+
```
SirPyTech: **typo**: This translation has been copy-pasted from another module, or it comes from previous attempts, anyway it shouldn't be here because right now there is no `field_res_company__free_shipping_product_default` in this module.
Please remove these translation files, or re-generate them for this module.

sale_coupon_default_product/models/coupon_program.py
```diff
@@ -0,0 +1,14 @@
+from odoo import api, models
+
+
+class CouponProgram(models.Model):
+    _inherit = "coupon.program"
+
+    @api.model
+    def create(self, vals):
+        # avoids creating multiple products if a coupon program is duplicated
```
SirPyTech: **suggestion**: if you want to edit the duplication flow, consider overriding `copy` (https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/odoo/models.py#L4701) or `copy_data` (https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/odoo/models.py#L4566) methods instead

SirPyTech (changes requested):
Thanks for the PR!
Please add a test that verifies the new behavior introduced by this module.

sale_coupon_default_product/readme/DESCRIPTION.rst
SirPyTech: :+1:

sale_coupon_default_product/models/coupon_program.py
SirPyTech: :+1:

sale_coupon_default_product/i18n/it.po
SirPyTech: :+1:

sale_coupon_default_product/models/coupon_program.py
SirPyTech: :+1:

sale_coupon_default_product/__manifest__.py
```diff
@@ -0,0 +1,14 @@
...
+    "name": "Sale coupon Default Product",
+    "summary": """avoids creating multiple discount products
+    if a coupon program is duplicated""",
+    "version": "14.0.1.0.0",
+    "category": "Hidden/Tools",
+    "author": "Ooops, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/sale-promotion",
+    "license": "LGPL-3",
+    "installable": True,
+    "depends": ["sale", "coupon", "sale_coupon_delivery"],
```
SirPyTech: **question**: why do we need all these dependencies?
I have only found the field `discount_line_product_id` that is being used in the `coupon` module: https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/coupon/models/coupon_reward.py#L51.

If only the `coupon` module is needed, please update this module's name and its references accordingly.

sale_coupon_default_product/models/coupon_program.py
```diff
@@ -0,0 +1,14 @@
...
+
+class CouponProgram(models.Model):
+    _inherit = "coupon.program"
+
+    def copy(self, default=None):
+        if default is None:
+            default = {}
+
+        if self.discount_line_product_id:
+            default["discount_line_product_id"] = self.discount_line_product_id.id
```
SirPyTech: **question**: Have you tried setting `copy=True` in the field definition?
Maybe the field's value wasn't being copied because the field explicitly has `copy=False` in its definition: https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/coupon/models/coupon_reward.py#L51.

If that works as expected, we could avoid this override entirely.

sale_coupon_default_product/models/coupon_program.py
```diff
@@ -0,0 +1,14 @@
...
+
+        return super(CouponProgram, self).copy(default)
```
SirPyTech: **suggestion**: 
```suggestion
        return super().copy(default=default)
```
If a function's (or in this case a method's) argument is declared with a keyword, it's better to use it in overrides so that if the original function has new arguments, we won't assign `default` to those.

I also removed the arguments of `super` because they aren't needed anymore.

sale_coupon_default_product/__manifest__.py
SirPyTech, line 1: **question**:
> Please add a test that verifies the new behavior introduced by this module.

(from https://github.com/OCA/sale-promotion/pull/290#pullrequestreview-3421453167)
News about the test?

SirPyTech (changes requested):

sale_coupon_default_product/__manifest__.py
Chionne27: Honestly, I prefer to leave the current name of the module to align it with the other module names in this repository.

sale_coupon_default_product/models/coupon_program.py
Chionne27: Your solution works well. Thanks for the suggestion

sale_coupon_default_product/models/coupon_program.py
Chionne27: I remove the method because i modified the `discount_line_product_id field `with `copy=False` in `coupon.reward` model

sale_coupon_default_product/__manifest__.py
Chionne27: I've done it. You can check

SirPyTech: 👍

sale_coupon_default_product/__manifest__.py
SirPyTech: The other modules are `sale_coupon...` because they depend on `sale_coupon`, if this module does not need such dependency, then it should be `coupon...`, like `coupon_limit`: https://github.com/OCA/sale-promotion/blob/477ede153d6ff791326916339e2729191bc2fcbf/coupon_limit/__manifest__.py#L13

sale_coupon_default_product/models/coupon_program.py
SirPyTech: :+1:

sale_coupon_default_product/models/coupon_program.py
SirPyTech: :+1:

SirPyTech (changes requested):

PR #292, opened 2025-08-14
[16.0][FIX] sale_loyalty_order_suggestion: do not apply promotions automatically

pilarvargas-tecnativa:
When there is a promotion of the type next_order_coupons, this generates a coupon for an order that has not yet been saved and therefore that coupon does not have the associated order that generated it. Furthermore, this line is unnecessary since promotions are applied automatically and, if the order already meets the conditions for a promotion to be applied, it will still be suggested in the wizard and the products added/required to apply the selected promotion will also be reported.

For these reasons, the line is removed as it applies redundant and unnecessary logic that slows down the process and generates useless coupons.

This has been detected and applied in the migration to v18.

@tecnativa 

@carlos-lopez-tecnativa @pedrobaeza please review

### DIFF
--- a/sale_loyalty_order_suggestion/tests/test_sale_loyalty_order_suggestion.py
+++ b/sale_loyalty_order_suggestion/tests/test_sale_loyalty_order_suggestion.py
@@ -1,5 +1,6 @@
 # Copyright 2023 Tecnativa - Pilar Vargas
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import Command
 from odoo.tests.common import Form, TransactionCase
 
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Any test that exercises the problem?

pilarvargas-tecnativa:
> Any test that exercises the problem?

image

sale_loyalty_order_suggestion/tests/test_sale_loyalty_order_suggestion.py
```diff
@@ -148,3 +148,44 @@ def test_02_suggested_promotion_for_product_a_auto_applicable(self):
...
+        next_order_coupon = self.env["loyalty.program"].create(
+            {
+                "name": "Test Loyalty Next Order Coupons",
+                "program_type": "next_order_coupons",
+                "trigger": "auto",
+                "applies_on": "future",
+                "rule_ids": [
+                    (
+                        0,
+                        0,
```
pedrobaeza:
```suggestion
                    Command.create(
```

pedrobaeza (commented):
Isn't this test very fragile if `sale_loyalty_auto_refresh` is installed first?

pilarvargas-tecnativa:
> Isn't this test very fragile if `sale_loyalty_auto_refresh` is installed first?

The autorefresh runs when the order is created, i.e. it is already saved. In this case, coupons will be created with an associated order. The problem that arises here is that when updating promotions before the order has been created and saved, a coupon was being created without an associated order, and this behaviour is not correct.

The test consists of checking that there are no coupons before the order is saved.

https://github.com/OCA/sale-promotion/blob/16.0/sale_loyalty_auto_refresh/models/sale_order.py#L29

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

PR #293, opened 2025-08-20
[18.0][MIG] website_sale_loyalty_suggestion_wizard: Migration to version 18.0

pilarvargas-tecnativa:
@Tecnativa TT54533

@carlos-lopez-tecnativa @CarlosRoca13 please review


website_sale_loyalty_suggestion_wizard/README.rst
```diff
@@ -0,0 +1,112 @@
...
+
+1. Go to /promotions and choose the one apply.
+2. You'll be prompted to check the promotion configuration.
+3. Apply after that.
+
+Option 2:
+
+1. Add a product in the promotion rules.
+2. A cart suggestion will appear.
+3. Add the promotion and configure the options.
```
carlos-lopez-tecnativa, line 55: I cannot test this option. Could you confirm whether this option is functional or not?

carlos-lopez-tecnativa (commented):
I cannot edit a rule, I get this error. I believe it is not related to this module but rather to the comment in the module `loyalty_criteria_multi_product`:
https://github.com/OCA/sale-promotion/pull/252#discussion_r2276448847

I think it is necessary that the module `loyalty_criteria_multi_product` itself adds the security rules, instead of relying on `sale_loyalty_criteria_multi_product`. Other modules like this one may have problems if `sale_loyalty_criteria_multi_product` is not installed. Please reconsider adding the security rights, at least READ access for internal users. 

@pedrobaeza What do you think?

image

pedrobaeza:
OK about moving the record rules.

website_sale_loyalty_suggestion_wizard/README.rst
pilarvargas-tecnativa: 1.
image

2.
image

3.
image

pilarvargas-tecnativa:
> I cannot edit a rule, I get this error. I believe it is not related to this module but rather to the comment in the module `loyalty_criteria_multi_product`: [#252 (comment)](https://github.com/OCA/sale-promotion/pull/252#discussion_r2276448847)

At the time, the modules in v15 were split, separating loyalty (coupon) from sale_loyalty (sale_coupon). It makes no sense to have a base loyalty module installed without the corresponding sale_* module, and Odoo's permissions must also be consistent with the model. Permissions are set in the sale_* modules, just as Odoo does.

https://github.com/odoo/odoo/blob/18.0/addons/loyalty/security/ir.model.access.csv

https://github.com/odoo/odoo/blob/18.0/addons/sale_loyalty/security/ir.model.access.csv

Furthermore, having permissions to establish a rule without the corresponding logic to apply it in an order leads to errors.

carlos-lopez-tecnativa (approved):
> At the time, the modules in v15 were split, separating loyalty (coupon) from sale_loyalty (sale_coupon). It makes no sense to have a base loyalty module installed without the corresponding sale_* module

Ok I understand you point, and has sense install the module `loyalty_criteria_multi_product` with the complementary module `sale_loyalty_criteria_multi_product`, as runboat all modules are installed but the `sale_loyalty_criteria_multi_product` is not merge yet, this complicate my test, but anyway, this error is not related to this PR, so go to ahead.

> Odoo's permissions must also be consistent with the model. Permissions are set in the sale_* modules, just as Odoo does.

I dont agree with Odoo to add a access rigth with 0,0,0,0 in the CRUD is equals to added nothing, but i think this is done to prevent the warning if a model has not access rigth :confused:

CarlosRoca13 (approved):

PR #298, opened 2025-08-26
[18.0][MIG] website_sale_loyalty_suggestion_wizard_multi_gift: Migration to version 18.0

pilarvargas-tecnativa:
@Tecnativa TT54534

@carlos-lopez-tecnativa @CarlosRoca13 please review


pedrobaeza:
/ocabot migration website_sale_loyalty_suggestion_wizard_multi_gift

website_sale_loyalty_suggestion_wizard_multi_gift/readme/USAGE.md
```diff
@@ -0,0 +1,19 @@
...
+
+1.  Add a product in the promotion multi-product rules.
+2.  A cart suggestion will appear.
+3.  Add the promotion and configure the options.
+
+Option 3:
+
+1.  Select a suggested promotion in the shopping cart from a list of
+    "Suggested promotions".
+2.  Add the promotion and configure the options.
```
carlos-lopez-tecnativa, line 19: Does this module depend on another module that you need to reference in `test-requirements`?

I tried to test it, but I couldn’t test it successfully. Please let me know if I missed anything.

image

I only see this option, but it looks like the native one.

image

I expected to see something like this in v16.

image

image

image

pilarvargas-tecnativa: Promotions are only suggested when the criteria for applying the rules require at least one product. I see that your rule is set for a minimum amount, so no products need to be added to the cart to obtain the reward.

carlos-lopez-tecnativa: Ok, thanks for pointing out my mistake. When the Minimum Quantity is zero, the promotion is not applied, and zero is the default value.

image

But why does Odoo standard suggest this promotion ( with minimum quantity = 0)? 🤔
image

carlos-lopez-tecnativa (approved):
Tested and working fine, LGTM

pilarvargas-tecnativa:
> But why does Odoo standard suggest this promotion ( with minimum quantity = 0)? 🤔

The promotion is applicable if the amount exceeds 50. It's not suggesting it; you already meet the requirement for the promotion to be applicable. If you go to the order in the backend, you will be able to apply the promotion. I am not really sure if that is covered by the base modules in the same way it is applied to orders. I would have to test it.

carlos-lopez-tecnativa:
> > But why does Odoo standard suggest this promotion ( with minimum quantity = 0)? 🤔
> 
> The promotion is applicable if the amount exceeds 50. It's not suggesting it; you already meet the requirement for the promotion to be applicable. If you go to the order in the backend, you will be able to apply the promotion. I am not really sure if that is covered by the base modules in the same way it is applied to orders. I would have to test it.

The `minimum purchase` requirement is satisfied, I was not referring to this variable. The `minimum quantity` is the one that prevents the promotion from being applied.
I changed the minimum quantity to 1 and now it works.

However, Odoo still suggested the promotion even when the minimum quantity = 0. In this module, **I expected that if the minimum quantity = 0, it would be considered as no restriction, and only if I set a minimum quantity greater than 0 should this condition be applied**. Why is it not working the same way as in Odoo base?

pilarvargas-tecnativa:
> However, Odoo still suggested the promotion even when the minimum quantity = 0. In this module, **I expected that if the minimum quantity = 0, it would be considered as no restriction, and only if I set a minimum quantity greater than 0 should this condition be applied**. Why is it not working the same way as in Odoo base?

I think I understand what you mean. Well, the wizard that opens when you suggest a promotion allows you to select the products to add. It is rare for the domain to be [] to add X units, and if so, it shows the complete list of products (which can be enormous). This is precisely the situation we find ourselves in with the minimum amount; the products to be displayed for addition would be ALL of them in order to reach the amount. That is an expandable part of the module, but it is not the main requirement. The requirement to be met is that if I purchase A and B, I receive C for free, and I add C to the cart, it suggests that there is a promotion for buying A and B.

CarlosRoca13 (approved):

PR #275, opened 2025-03-28
[18.0][MIG] loyalty_coupon_no_mail: Migration to 18.0

natuan9:
[Screencast from 2025-03-26 10-13-58.webm](https://github.com/user-attachments/assets/e45478a9-950c-4461-9b53-edb4fae6bd4c)

PR #301, opened 2025-08-27
[18.0][MIG] loyalty_coupon_no_mail: Migration to 18.0

xaviedoanhduy:
Supersedes: https://github.com/OCA/sale-promotion/pull/275

In addition, this PR also adds test cases: https://github.com/OCA/sale-promotion/pull/301/commits/c05cd438651a1d374d51bf8bcdd0b0cd7da3d643


xaviedoanhduy:
hello @pedrobaeza, @wlin-kencove, @kobros-tech 👋
could you please take another look? I've addressed the change requests in the latest updates. Thanks! 🙏

pedrobaeza:
The last commit tag should be `[IMP]`, not `[ADD]`, as you are not adding a module.

pedrobaeza:
I'm also intrigued by the `[FIX]` one. Why being split and steps to reproduce the problem?

xaviedoanhduy:
> I'm also intrigued by the [FIX] one. Why being split and steps to reproduce the problem?

ah sorry, maybe my original intention was to separate them so we could discuss them more easily - I merged them into the `MIG` commit where they should belong

kobros-tech (approved):
all right

pedrobaeza (changes requested):
Well, the tests I expect is that when a coupon is granted, no mail is generated.

xaviedoanhduy:
hi @pedrobaeza, 
well i added 2 test cases about sent mail or not
is it ok for you now?

pedrobaeza (approved):
/ocabot merge nobump

PR #231, opened 2024-12-14
[18.0][MIG] loyalty_partner_applicability

sbejaoui:

PR #305, opened 2025-08-29
[18.0][MIG] loyalty_partner_applicability: Migration to version 18.0

pilarvargas-tecnativa:
supersede https://github.com/OCA/sale-promotion/pull/231

@Tecnativa TT54449

@victoralmau @carlos-lopez-tecnativa @lmignon please review


lmignon:
~~The restriction seems to be not taken into account into the UI. If I set a specific partner on Loyalty program and click on the reward button on a so for another partner, this partner benefits from this specific loyalty program~~

My bad, we need the `sale_loyalty_partner_applicability` addon to test the current addon

loyalty_partner_applicability/models/loyalty_program.py
```diff
@@ -0,0 +1,105 @@
...
+    def _get_eval_partner_domain(self):
+        self.ensure_one()
+        return safe_eval(
+            self.partner_domain,
+            {"datetime": datetime, "context_today": datetime.datetime.now},
+        )
+
+    def _is_coupon_sharing_allowed(self):
+        allow_sharing = (
+            self.env["ir.config_parameter"].sudo().get_param("allow_coupon_sharing")
```
carlos-lopez-tecnativa, line 44: I think this parameter should be documented in the README to clarify its purpose. Otherwise, I would only know that this parameter exists by reading the code.

Update: It seems this was added in the README here: https://github.com/OCA/sale-promotion/blob/16.0/sale_loyalty_partner_applicability/readme/CONFIGURE.rst

carlos-lopez-tecnativa (approved):
Code review LGTM

carlos-lopez-tecnativa:
I would be interested in reviewing this module once the other module https://github.com/OCA/sale-promotion/blob/16.0/sale_loyalty_partner_applicability is migrated.

lmignon (approved):
LGTM Code review

PR #308, opened 2025-09-24
[16.0][FIX] sale_loyalty_multi_gift: Update translation files

pilarvargas-tecnativa:
@Tecnativa TT58161

@pedrobaeza @victoralmau please review


pedrobaeza (approved):
As this requires the cleaning of the .pot file, let's do it here and not in Weblate.

/ocabot merge patch

PR #319, opened 2026-03-16
[18.0][FIX] sale_loyalty_multi_gift: avoid singleton error when multiple rewards exist

pilarvargas-tecnativa:
Ensure a single reward before accessing its fields to prevent singleton errors when more than one reward is returned.

@Tecnativa TT61527

@victoralmau @carlos-lopez-tecnativa please review

### DIFF
--- a/sale_loyalty_multi_gift/models/sale_order.py
+++ b/sale_loyalty_multi_gift/models/sale_order.py
@@ -15,31 +15,28 @@ class SaleOrder(models.Model):
         self.ensure_one()
         self._update_programs_and_rewards()
         claimable_rewards = self._get_claimable_rewards()
-        if (
-            len(claimable_rewards) == 1
-            and claimable_rewards.get(next(iter(claimable_rewards))).reward_type
-            == "multi_gift"
-            and any(
-                len(record.reward_product_ids) > 1
-                for record in claimable_rewards.get(
-                    next(iter(claimable_rewards))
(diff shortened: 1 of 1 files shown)


sale_loyalty_multi_gift/models/sale_order.py
```diff
@@ -15,31 +15,31 @@ def action_open_reward_wizard(self):
...
-            return {
-                "type": "ir.actions.act_window",
-                "view_mode": "form",
-                "res_model": "sale.loyalty.reward.wizard",
-                "target": "new",
-                "context": ctx,
-            }
-        else:
-            return super().action_open_reward_wizard()
+        if len(claimable_rewards) == 1:
```
pedrobaeza, line 18: What happens if `len(claimable_rewards)` != 1?

pilarvargas-tecnativa: image
I had missed an indentation.

If there is more than one promotion, open the wizard to select it

pedrobaeza: But now there's another hanging `if` inside...

pilarvargas-tecnativa: solved, thanks

pedrobaeza (approved):

victoralmau (approved):

PR #320, opened 2026-03-18
[18.0][FIX] website_sale_loyalty_suggestion_wizard, sale_loyalty_order_suggestion: fix website promotion wizard issues

pilarvargas-tecnativa:
This PR fixes two issues affecting the website loyalty promotion wizard.

In module website_sale_loyalty_suggestion_wizard the quantity onchange handler was iterating a jQuery collection using forEach, which is not supported and raises a Javascript error. The iteration has been replaced with each, which is the correct way to iterate jQuery collections.

In module sale_loyalty_order_suggestion the overridden action_apply method was calling _apply_loyalty_rule_lines_to_order before delegating to the parent implementation. Since the parent method already performs this operation, the rule lines were applied twice when using the website promotion wizard. The fix calls the parent method with skip_apply_loyalty_rule_lines=True to prevent the duplicate execution.

@Tecnativa TT61563

@victoralmau @pedrobaeza please review

### DIFF
--- a/sale_loyalty_order_suggestion_multi_gift/wizard/sale_loyalty_reward_wizard.py
+++ b/sale_loyalty_order_suggestion_multi_gift/wizard/sale_loyalty_reward_wizard.py
@@ -6,7 +6,10 @@ class SaleLoyaltyRewardWizard(models.TransientModel):
 
     def action_apply(self):
         self._apply_loyalty_rule_lines_to_order()
-        super().action_apply()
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

david-banon-tecnativa (approved):

victoralmau (approved):
The changes appear to be consistent.

PR #321, opened 2026-04-30
[18.0][ADD] loyalty_margin_computation: New module

Andrii9090-tecnativa:
It is a base module that only adds a field to `loyalty.reward` to define a formula for margin calculation

@pedrobaeza can you review please?

TT62183
@Tecnativa


loyalty_margin_computation/views/loyalty_reward_views.xml
```diff
@@ -0,0 +1,40 @@
...
+                        <div
+                            class="alert alert-info"
+                            role="alert"
+                            name="variables_to_use"
+                        >
+                            <p>The following variables can be used:</p>
+                            <ul>
+                                <li><code
+                                >env</code>: The Sale Order Line environment</li>
+                                <li><code>reward</code>: The reward record</li>
```
pedrobaeza, line 26: Only this variable is available/needed?

loyalty_margin_computation/views/loyalty_reward_views.xml
```diff
@@ -0,0 +1,40 @@
...
+                            </ul>
+                            <div>
+                                <p>Example of Python code:</p>
+                                <pre
+                                    style="white-space: pre-wrap"
+                                >result = 10 if list_price > 100 else 5</pre>
```
pedrobaeza: This example doesn't seem correct according the stated variables.

loyalty_margin_computation/models/loyalty_reward.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2026 Tecnativa - Andrii Kompaniiets
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class LoyaltyReward(models.Model):
+    _inherit = "loyalty.reward"
+
+    sale_margin_formula = fields.Text(
```
pedrobaeza, line 10: Isn't there a check to see if the syntax is correct?

loyalty_margin_computation/views/loyalty_reward_views.xml
Andrii9090-tecnativa: It's a technical base to be extended with top level functionality and more variables adds other module

pedrobaeza:
The same: it should go into OCA/margin-analysis, as it's more related to analysis and the base module is there.

PR #326, opened 2026-05-07
[19.0][MIG] loyalty_program_chatter: Migration to 19.0

BhaveshHeliconia:


kalpeshgajera-uncannycs (changes requested):
A few minor points to consider before merge:

1. **Redundant `_name`**: In `models/loyalty_program.py`, `_name = "loyalty.program"` is redundant when extending the model. You can keep just `_inherit = ["loyalty.program", "mail.thread", "mail.activity.mixin"]`.
2. **Tests**: Consider adding basic tests to verify chatter/activity creation on `loyalty.program` per OCA migration guidelines.
3. **i18n headers**: `.pot` and `.po` files still reference older versions (16.0 / 18.0) and should be updated to `19.0`.

BhaveshHeliconia:
Thanks for the review, @kalpeshgajera-uncannycs! I’ve addressed the other points. I initially removed the explicit _name, but that caused a CI failure with the Odoo 19 warning, so I’ve reverted that change. I don’t think additional tests are needed here since there is no custom logic to test, and the i18n headers are handled by the Weblate bot.

Also, this review looks AI-generated and quite like a copy-paste from an AI response. Please make sure to review and verify the comments before posting them, as mentioned in the OCA AI policy:
https://github.com/OCA/.github/blob/master/AI_POLICY.md

PR #340, opened 2026-05-20
[18.0][FIX] sale_loyalty_order_suggestion_multi_gift preserve same-product paid qty

pilarvargas-tecnativa:
Do not reduce existing order lines when the required product is the same as the selected reward product. In that case, the reward must be added as a separate free line on top of the purchased quantity.

For different reward products already present in the order, convert the existing paid quantity into reward quantity by either reducing the line or removing it completely when the remaining quantity would be zero.

@Tecnativa TT62612

@CarlosRoca13 @victoralmau please review


victoralmau (approved):
The changes seem to make sense.

CarlosRoca13 (approved):

PR #341, opened 2026-05-22
[18.0][FIX] loyalty_criteria_multi_product: Clear the minimum quantity field

pilarvargas-tecnativa:
When modifying the multi-product rule, you must also clear the minimum quantity field so that it does not interfere with the promotion rules if it has been set previously.

@Tecnativa TT62602

@victoralmau @carlos-lopez-tecnativa please review


pedrobaeza (commented):
This is OK for data consistency, but shouldn't the module checking that minimum_qty ignore the field if `loyalty_criteria` is `multi_product`?

pilarvargas-tecnativa:
> This is OK for data consistency, but shouldn't the module checking that minimum_qty ignore the field if `loyalty_criteria` is `multi_product`?

The real problem is that Odoo evaluates all the standard criteria together within `_program_check_compute_points()`

https://github.com/odoo/odoo/blob/95b29dc6dc6c094789282ad32fb6a6cacda88c7f/addons/sale_loyalty/models/sale_order.py#L1163

There is no clear point where `minimum_qty` is evaluated on its own, so that we can ignore only that criterion in `multi_product` without rewriting part of the standard logic.

pedrobaeza:
OK, but then this should include a migration script that puts 0 on all the existing criteria for multi_product. Even more, you are acting on the onchange, but when changing things on code, the inconsistency will appear the same. You need a more reliable method to keep things consistent.

loyalty_criteria_multi_product/migrations/18.0.1.0.1/post-migration.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env["loyalty.rule"].search([("loyalty_criteria", "=", "multi_product")]).write(
+        {
+            "minimum_qty": 0.0,
+        }
```
carlos-lopez-tecnativa:
```suggestion
        {
            "minimum_qty": 0.0
        }
```

loyalty_criteria_multi_product/models/loyalty_rule.py
```diff
@@ -20,14 +20,32 @@ class LoyaltyRule(models.Model):
         inverse_name="rule_id",
     )
 
+    def _get_multi_product_reset_values(self):
+        return {
+            "minimum_qty": 0.0,
+            "minimum_amount": 0.0,
+            "product_domain": False,
+            "product_ids": [(5, 0, 0)],
```
carlos-lopez-tecnativa:
```suggestion
            "product_ids": [Command.clear()],
```

carlos-lopez-tecnativa (approved):
LGTM

loyalty_criteria_multi_product/models/loyalty_rule.py
pilarvargas-tecnativa: thanks

pedrobaeza (approved):
/ocabot merge nobump

PR #359, opened 2026-08-25
[18.0][FIX] loyalty_incompatibility: don't spread incompatibilities between programs

### LINKED ISSUE #336: loyalty_incompatibility configuration issue
## Module
loyalty_incompatibility @ sale_loyalty_incompatibility

## Describe the bug
When adding an Incompatible Program to a New program, if that Incompatible program is listed on any other programs, it then automatically adds any of the incompatible programs from that program, to this new program. And you are then unable to remove any of them from any of those programs. The only way to remove them is to just start over and create a whole new promotion.

## To Reproduce
V18

1. Create a Promotion and in the Incompatible field select 2 other Promotions that you do not want to work with this promotion
2. Create another new Promotion and in the Incompatible field select one of those promotions that you had selected in the previous promotion and click save.  Upon save, even though you only selected one of those, it automatically adds the other one as well
3. Try to click back into that promotion to remove one of those incompatible promotions and click save, it adds it right back. So you are unable to remove them. Even if you try to go back to the original one that you created and remove the Incompatible promotions, once you hit save, it brings them back in. 

**Expected behavior**
One would need the ability to select some Incompatible promotions for one promotion, but then select different Incompatible promotions for another promotion, but this cannot be done with this module. 

**Additional context**
Please watch this Video for context: https://www.awesomescreenshot.com/video/52560274?key=a5f7eb7c70c5b0fd658923e63ce34aaa

kaynnan:
Fixes #336.

Selecting one incompatible program silently pulls in the others it was already incompatible with. With A ⊥ {B, C}, setting D ⊥ {B} leaves D incompatible with both B and C.

The inverse keeps the relation symmetric with:

```python
program.incompatible_promotion_ids.incompatible_promotion_ids |= program
```

`program.incompatible_promotion_ids` is usually more than one record, and reading an x2m over a multi-record set returns the union of their values, which then gets written back to every one of them. So B and C, which only had A in common, end up carrying each other's incompatibilities.

Each counterpart is now updated on its own with a link/unlink command, so nothing leaks between them. The recursive call the write triggers on the counterpart is skipped through the context, as the sync only needs to run for the program that was actually written.

Added a test reproducing the report, plus one covering symmetric removal, which had no coverage.

### DIFF
--- a/loyalty_incompatibility/models/loyalty_program.py
+++ b/loyalty_incompatibility/models/loyalty_program.py
@@ -1,7 +1,7 @@
 # Copyright 2021 Tecnativa - David Vidal
 # Copyright 2023 Tecnativa - Stefan Ungureanu
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import Command, fields, models
 
 
 class LoyaltyProgram(models.Model):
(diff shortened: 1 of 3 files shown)


kaynnan:
Ping @shopakr

